PR #44, opened 2015-04-23
Ansible deployments

powellc:
A really simple playbook for setting up pypicloud using ansible. I can docs if you have any interest in this.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,3 +40,8 @@ pypicloud_env
 _build
 _local.db
 *.ini
+ansible/group_vars/local
+ansible/group_vars/production
+ansible/group_vars/staging
+ansible/hosts
+.vagrant
--- a/ansible/group_vars/all
+++ b/ansible/group_vars/all
@@ -1,0 +1,7 @@
+---
+hostname: pypi
+pypicloud_version: 0.2.10
+timezone: America/New_York
+locale: en_US.UTF-8
+deployment_name: pypicloud
+port_number: 6543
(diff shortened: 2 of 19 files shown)


stevearc:
This is great! Question: do projects typically include ansible playbooks in the same repository? My instinct would be to put it in a separate repo so the versions aren't tied together, but I'm willing to go with whatever the common practice is. Also, it seems like there's a bunch of stuff in there that's unrelated to pypicloud (all the UFW commands, the `info.html` page, etc). For anything in the main pypicloud repo I'd prefer it to be trimmed down to just the relevant bits, to make it easier to skim and twiddle. Thoughts?

powellc:
Hmm, good question. Some projects break it out, and some drop them in. As I think in it, it might be better to have an ansible-pypicloud repository. That would also explain the reason for things likw UFW and what-not.

Basically you can point the ansible script at a vanilla Ubuntu/Debian VPS and it will setup a running pypicloud instance. In order to do that, there are some housekeeping things Ansible needs to do like set timezones properly, and setup a firewall.

I'll setup a parallel repository on Github for the ansible scripts and you can close this baby down. References to the ansible build are welcome too :)

powellc:
Here we go: https://github.com/powellc/ansible-pypicloud

stevearc:
Great! I added a reference [in the docs](https://github.com/mathcamp/pypicloud/commit/c63af0fd128b849db5a23408f0c508fb51b5db89) that will get pushed out once I cut a new release (coming Soon&trade;)

powellc:
Awesome. Thanks for a great project!

CP

Issue #42, opened 2015-02-26
repository failed to get version higher than current stored version in repo


avihoo:
Hi,

I encountered an issue, when I have for example, application called app with version 0.1 on my pypicloud repository, and I'm asking to download app with version other than 0.1 that does not exists on the pypi repo, instead of fetching it from the world's pypi repo, I get an error saying the package is not found.

Using pypicloud version 0.2.7 with fallback = cache and S3 storage.

Thanks.

johtso:
It seems that if there is *any* version of a package cached, then pypicloud does not fetch the different version of a package if you request it.

This just bit us too.

stevearc:
This is somewhat by design, and somewhat a limitation of pip. Let's take the following case:

<table>
  <thead>
    <tr>
      <th>pypicloud</th>
      <th>world pypi</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td>mypackage 0.1</td>
      <td>mypackage 0.1</td>
    </tr>
    <tr>
      <td></td>
      <td>mypackage 0.2</td>
    </tr>
  </tbody>
</table>

When you do a `pip install mypackage`, the desired behavior is for it to install version `0.1`. If you're deploying software to internal servers, you don't want your library versions to change just because the maintainer released a new version upstream, you want consistency. In this case, pip requests the list of packages, pypicloud says "I have 0.1", and pip selects 0.1 as the best candidate.

But what happens when you do `pip install mypackage>0.1`? Pip doesn't tell pypicloud that it needs a version *other* than 0.1. It doesn't tell pypicloud anything. All it does is request all package versions for `mypackage`, and then pip handles the version selection. This is the problem you've encountered. Unfortunately there is no way for pypicloud to handle this without breaking the above behavior.

There are a few options for where to go from here. There should be a "fetch" button on the web interface that will allow you to grab additional packages from upstream. You can also use the [http api](http://pypicloud.readthedocs.org/en/latest/topics/api.html#post-api-fetch) You can use this to acquire the packages you need, but it is a manual process (by design).

You can specify the fallback url with pip: `pip install --index-url http://my.pypicloud.com/simple/ --extra-index-url http://pypi.python.org/simple`. Apart from the obvious downside, this also means that if pypicloud doesn't have the desired package version, pip will fetch it directly from pypi and not cache it in pypicloud.

[...]

PR #45, opened 2015-04-27
Fix some problems when server is in cache mode

smokey42:
* Packages on PYPI with older versions which have invalid PEP440 versions
   would crash distlib in default mode

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@
 *.egg-info
 dist
 build
+.eggs
 eggs
 parts
 bin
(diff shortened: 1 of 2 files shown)


stevearc:
First two commits are great, definitely will merge. Third commit requires a bit more discussion. Check out #42 for details, but the short version is that the current behavior of `cache` is a feature. It seems like people want the other behavior, but I'd like to put it behind a different mode called `mirror` or something. You can do this if you like, or I can look into it when I have time.

smokey42:
Removed the last commit. Still thinking about the other issue. There doesn't seem to be a particularly clean solution for this problem though.

smokey42:
I just realized I don't need the other fix, because fallback=redirect is ok with me.

stevearc:
Awesome, thanks! I'll cut a new release sometime soon.

PR #47, opened 2015-05-08
Correct time comparison in s3.py, get_url

SofiaMargariti:
As it is now, in the case of temporary credentials, there is an inconsistent comparison between `seconds` which has a unix timestamp value and `expire_after` which is an arbitrary number of seconds. As a result `if seconds < expire_after:` always returns `False` and the expiration time does not get updated correctly. This commit fixes that (cc @stevearc).


stevearc:
Great find! Thanks for the PR!

PR #49, opened 2015-05-25
Fixed fallback mirror mode

avihoo:
Bug fixed when private package needs to be download and not found in fallback url, the get_fallback_packages returns None and then there is an assignment for NoneType. fixed with returning empty dict in case None has returned (usually for private packages not stored in fallback url's which are mostly wide)

### DIFF
--- a/pypicloud/views/simple.py
+++ b/pypicloud/views/simple.py
@@ -65,7 +65,7 @@ def package_versions(context, request):
     pkgs = {}
     if fallback == 'mirror':
         if can_update_cache:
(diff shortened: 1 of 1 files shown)


stevearc:
Thanks for the PR! What do you think of 545c43355f4f8dafc31e782340ea88dbabaa1aa7 as a solution instead? (with 545c43355f4f8dafc31e782340ea88dbabaa1aa7 also to clean it up a bit more)

stevearc:
Great! Closing because fixed in the above commits.

PR #50, opened 2015-06-01
Add waitress as dependency to setup.py

longnguyen11288:


stevearc:
Hi Long, thanks for the PR! I actually intentionally excluded `waitress` from the list of dependencies because it's not technically part of the pypicloud package. It's not at all necessary to run pypicloud, it's just a convenient webserver that can serve the WSGI app for development or small deploys.

longnguyen11288:
Hmm ok, I wrote a blog post https://blog.starkandwayne.com/2015/06/01/running-your-own-pypi-server-on-cloud-foundry/ on running it on Cloud Foundry. I'll leave it as manual step to add it so it can run on Cloud Foundry

PR #51, opened 2015-06-09
bug fixed in redis cache, after uploading package from admin UI

avihoo:
Bug: On S3 storage with redis as cache server, when uploaded package from admin UI, the packages uploaded to S3 but are not shown in /simple

Solution: After uploading initiate the reload from storage.

### DIFF
--- a/pypicloud/views/api.py
+++ b/pypicloud/views/api.py
@@ -105,8 +105,12 @@ def download_package(context, request):
 def upload_package(context, request, content):
     """ Upload a package """
     try:
-        return request.db.upload(content.filename, content.file,
+        request.db.upload(content.filename, content.file,
                                  name=context.name)
+        response = request.response
+        request.db.reload_from_storage()
+        return response
+
     except ValueError as e:  # pragma: no cover
         return HTTPBadRequest(*e.args)
 


stevearc:
There should already be code that makes sure the cache is populated from an upload. It's the same code path whether you upload the package with `python sdist upload` or use the web UI. I'm not able to reproduce this issue on my machine. Is there anything odd about this package or setup that could be causing the issue? Non-standard version numbers? Unicode characters in the package name?

avihoo:
The package name contains dashes, for example: my-package-0.01.tar.gz
but when initiating the function request.db.reload_from_storage() it does appear on the packages list, and then available via pip install.
The cache server is local reddit server and storage is S3 based.

stevearc:
Still can't reproduce. Can you confirm that the data is missing from redis after the upload? From the redis-cli a simple `smembers pypicloud:set` should show you all distinct packages, and `smembers pypicloud:set:<package name>` should show you all files for a package. If the data is missing, the next step would be to instrument [this method](https://github.com/mathcamp/pypicloud/blob/e2470b87519729e7fff0abe180b5ad131cdad397/pypicloud/cache/redis_cache.py#L91) to see what's going wrong with the save.

avihoo:
OK, so it seems that I can't reproduce the problem again, and not it's working on every scenario I had problems with. Must have been problems with my cache server.

Sorry for all the mess, ticket can be closed and pull request as well.

Thanks!

PR #53, opened 2015-06-26
hooks exception formatting to allow multiline tracebacks in a single json log

a-tal:
currently tracebacks are split across multiple lines/log messages, this will keep them together

### DIFF
--- a/pypicloud/__init__.py
+++ b/pypicloud/__init__.py
@@ -100,7 +100,6 @@ def includeme(config):
                            cache_max_age=cache_max_age)
 
 
-
 def traceback_formatter(excpt, value, tback):
     """ Catches all exceptions and re-formats the traceback raised.
     """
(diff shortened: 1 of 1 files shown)


pypicloud/__init__.py
```diff
@@ -97,9 +100,27 @@ def includeme(config):
...
+
+    sys.stdout.write("".join(traceback.format_exception(excpt, value, tback)))
+
+
+def hook_exceptions():
+    """ Hooks into the sys module to set our formatter.
+    """
+
+    # reopen stdout in non buffered mode
+    sys.stdout = os.fdopen(getattr(sys.stdout, "fileno", lambda: 1)(), 'w', 0)
```
stevearc: I would prefer if `hook_exceptions` was a no-op if `sys.stdout` doesn't have a `fileno()`. As it's written, I think it might mess up test output formatting.

stevearc:
Looks good. One comment for the test case behavior.

stevearc:
Thanks!

PR #55, opened 2015-07-14
adds ldap auth plugin

a-tal:


stevearc:
Nice! I'll get this reviewed ASAP

a-tal:
cool, thanks. I have it working pretty well right now in a little test vm.
hit me up on chat if you want.

pypicloud/access/__init__.py
```diff
@@ -22,6 +23,8 @@ def includeme(config):
         dotted_name = RemoteAccessBackend
     elif dotted_name == 'sql':
         dotted_name = SQLAccessBackend
+    elif dotted_name == 'ldap' and LDAP_ENABLED:
+        dotted_name = LDAPAccessBackend
```
stevearc: If `ldap` isn't installed this will produce an unclear message (something about being unable to import the path). I recommend looking at how the [caching backend](https://github.com/mathcamp/pypicloud/blob/master/pypicloud/cache/__init__.py#L20) sets itself up (since it isn't guaranteed to have flywheel installed at runtime). It does a lazy import and [prints out a clear error message](https://github.com/mathcamp/pypicloud/blob/c289855f4251a816e94aacc7c52f07e82ba4aceb/pypicloud/cache/dynamo.py#L20) if it fails.

pypicloud/access/_ldap.py
```diff
@@ -0,0 +1,391 @@
+"""LDAP authentication plugin for pypicloud."""
+
+
+try:
+    import ldap  # pylint: disable=F0401
```
stevearc: Instead of the pylint disable flag, you can just put `python-ldap` in the `requirements_build.txt` file. That'll be required later anyway if any tests are written for it

pypicloud/access/_ldap.py
```diff
@@ -0,0 +1,391 @@
...
+            return False
+        else:
+            return True
+        finally:
+            LDAP._connect()
+
+
+class LDAPAccessBackend(IAccessBackend):
+    """
+    This backend allows you to store all user and package permissions in a SQL
```
stevearc: This is just incorrectly copy-pasted, right? I don't see any SQL here

pypicloud/access/_ldap.py
```diff
@@ -0,0 +1,391 @@
...
+
+    DEFAULT_PERMISSIONS = {
+        "admin": ("read", "write"),
+        "authenticated": ("read", "write"),
+        "everyone": ("read",),
+    }
+
+    def __init__(self, request=None, **kwargs):
+        self.request = request
+        super(LDAPAccessBackend, self).__init__(request, **kwargs)
```
stevearc: You don't need the `__init__` method. The base class `__init__` should do this

pypicloud/access/_ldap.py
```diff
@@ -0,0 +1,391 @@
...
+        ----------
+        username : str
+        password : str
+
+        Returns
+        -------
+        valid : bool
+            True if user credentials are valid, false otherwise
+
+        """
```
stevearc: I think you can exclude most of these lengthy docstrings if you want, since they're on the base class. Up to you.

pypicloud/templates/base.jinja2
```diff
@@ -9,7 +9,7 @@
   {%- block all_css -%}
     <link href="{{ 'pypicloud:static/css/lib/bootstrap-3.0.3.min.css' | static_url }}" rel="stylesheet">
     <link href="{{ 'pypicloud:static/css/lib/font-awesome-4.0.3/css/font-awesome.min.css' | static_url }}" rel="stylesheet">
-    <link href="{{ 'pypicloud:static/css/base.css' | static_url }}" rel="stylesheet"></link>
+    <link href="{{ 'pypicloud:static/css/base.css' | static_url }}" rel="stylesheet">
```
stevearc: Good catch

pypicloud/access/_ldap.py
```diff
@@ -0,0 +1,391 @@
...
+
+    @staticmethod
+    def _connect():
+        """Initializes the python-ldap module and does the initial bind."""
+
+        LDAP._server = ldap.initialize(LDAP._url)
+        LDAP._server.simple_bind_s(LDAP._service_dn, LDAP._service_password)
+
+    @staticmethod
+    @reconnect
```
stevearc: What's the reasoning behind which methods you wrap in `@reconnect`? Why do you not wrap every call?

pypicloud/access/_ldap.py
```diff
@@ -0,0 +1,391 @@
...
+                if LDAP._id_field in result[1]:
+                    LDAP._all_users[result[1][LDAP._id_field][0]] = result[0]
+
+        return list(LDAP._all_users.values())
+
+    @staticmethod
+    def all_usernames():
+        """Returns a list of all user names."""
+
+        return list(LDAP._all_users.keys())
```
stevearc: It seems like you're relying on `all_usernames()` only ever being called after `all_users()`. Is that guaranteed to be true? It's not obvious to me just from looking at how it's called.

pypicloud/access/_ldap.py
```diff
@@ -0,0 +1,391 @@
...
+    """
+    This backend allows you to store all user and package permissions in a SQL
+    database and authenticate against a LDAP server
+    """
+
+    DEFAULT_PERMISSIONS = {
+        "admin": ("read", "write"),
+        "authenticated": ("read", "write"),
+        "everyone": ("read",),
+    }
```
stevearc: In general I would prefer that the user configure this with the `pypi.default_read` and `pypi.default_write` options. Unless I misunderstand, this makes it completely inflexible, so if someone wants logged-out users to be unable to download packages, there is no way to do that. Is there a good reason to do the configuration this way?

stevearc:
Good stuff! Left a few comments.

pypicloud/access/_ldap.py
a-tal: ok. will do

pypicloud/access/__init__.py
a-tal: this should be clearer once python-ldap is in the requirements.txt

pypicloud/access/_ldap.py
a-tal: yeah, it is. will fix

pypicloud/access/_ldap.py
a-tal: yeah, they're a bit long, but :+1: was using them as a guide. thanks for writing the initial ones!

pypicloud/access/_ldap.py
a-tal: basically any method that deals with the LDAP._server object, besides the initial one as we can be pretty sure it's not going to die. ldap connections are usually fairly short lived though, so just to auto rebind whenever that happens

pypicloud/access/_ldap.py
a-tal: yeah. that's the case currently because the call in `configure`. the side effect is that your ldap cache is built as soon as the app is started, but I'm starting to lean back the other way of building the cache on first request.

I'll change this up to be more clear and loading later

pypicloud/access/_ldap.py
a-tal: good catch, will change that

a-tal:
:bell: hey, any outstanding issues left on this?

setup.py
```diff
@@ -28,6 +28,7 @@
     'six',
     'transaction',
     'zope.sqlalchemy',
+    'python-ldap',
```
stevearc: I actually preferred what you were doing earlier with `extras_require`. It would be better to not require every installation to install `python-ldap`

pypicloud/access/_ldap.py
```diff
@@ -0,0 +1,292 @@
...
+        super(LDAPAccessBackend, self).__init__(request, **kwargs)
+        self.group_map = group_map
+
+    @classmethod
+    def configure(cls, settings):
+        kwargs = super(LDAPAccessBackend, cls).configure(settings)
+        LDAP.configure(settings)
+        kwargs["group_map"] = {
+            "admin": ("read", "write"),
+            "authenticated": ("read",),
```
stevearc: I don't think you need these two lines anymore, since the `settings.get()` below handles all the behavior you want

stevearc:
Two last small nits. Thanks for putting so much work into this!

Fixes #16

pypicloud/access/_ldap.py
a-tal: they're usually not needed yeah. but if someone configures default_write to some other group, without those lines, admin would lose write access. same with authenticated and the inherent default read.

stevearc:
:+1:

PR #60, opened 2015-10-12
Update the pypi fallback url from http://pypi.python.org/simple to https://pypi.python.org/simple .

dbaxa:


stevearc:
Fantastic :+1:

PR #63, opened 2015-12-02
waitress is required, so add "pip install" to quickstart

philwhln:


stevearc:
This is what the [server] is for in `pip install pypicloud[server]`. The [extras_require in setup.py](https://github.com/mathcamp/pypicloud/blob/master/setup.py#L94) should make that command also install waitress. Did you run that command and *not* end up with waitress installed?

PR #64, opened 2016-01-08
Add view for all packages

pkariz:
currently we need to define link for every package in pypicloud separately in buildout.cfg, example:
find-links +=
    https://whatever.herokuapp.com/simple/package1
    https://whatever.herokuapp.com/simple/package2

with this change you can have cleaner buildout config files which is very nice when you have many packages, example:
find-links +=
    https://whatever.herokuapp.com/packages

### DIFF
--- a/pypicloud/views/packages.py
+++ b/pypicloud/views/packages.py
@@ -3,7 +3,7 @@ from pyramid.view import view_config
 from pyramid_duh import addslash
 
 from pypicloud.route import PackagesResource
-from pypicloud.views.simple import _packages_to_dict
+from pypicloud.views.simple import packages_to_dict
 
 
 @view_config(context=PackagesResource, request_method='GET', subpath=(),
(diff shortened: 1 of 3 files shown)


pypicloud/views/packages.py
```diff
@@ -0,0 +1,21 @@
+""" View for cleaner buildout calls """
+from pyramid.view import view_config
+from pyramid_duh import addslash
+
+from pypicloud.route import PackagesResource
+from pypicloud.views.simple import _packages_to_dict
```
stevearc: Small nit: since this function is no longer local to the file where it's declared it should be renamed to `packages_to_dict`

stevearc:
One tiny request, but otherwise looks good. Thanks for including all the docstrings and tests!

pkariz:
Agree with you, fixed that. Docstrings and tests are as important as other code, so np :)

stevearc:
Released in 0.3.7
Thanks for the PR!

PR #69, opened 2016-01-29
Revamp dynamo's reload_from_storage

pkoch:
Still a bit rough on some edges, but this is probably a good place to start reviewing.

There's decisions marked with XXX that I'm not sure of. I'd appreciate your thoughts on those.

There's also two problems in flywheel, marked with HACK, that I had to dance around. These should probably be fixed there.

I've experimented with most of the cases I could think of, but, so far, I only coded the ones I don't need precise timings for (I've been doing those manually with ipdb).

Let me know if there's anything that's terribly off base, or if there's some nip/tucking I should apply. Meanwhile, I'm going to proceed to concoct a way to test the consistency clashes.

### DIFF
--- a/pypicloud/cache/base.py
+++ b/pypicloud/cache/base.py
@@ -8,7 +8,7 @@ from pyramid.settings import asbool
 import posixpath
 from pypicloud.models import Package
 from pypicloud.storage import get_storage_impl
-from pypicloud.util import parse_filename, normalize_name
+from pypicloud.util import parse_filename, normalize_name, EPOCH
 
 
 LOG = logging.getLogger(__name__)
@@ -194,7 +194,7 @@ class ICache(object):
                 'name': name,
                 'stable': None,
                 'unstable': '0',
-                'last_modified': datetime.fromtimestamp(0),
+                'last_modified': EPOCH,
             }
             for package in self.all(name):
                 if not package.is_prerelease:
--- a/pypicloud/cache/base.py
+++ b/pypicloud/cache/base.py
@@ -66,7 +66,14 @@ class ICache(object):
         return self.storage.download_response(package)
 
     def reload_from_storage(self):
-        """ Make sure local database is populated with packages """
+        """
+        Make sure local database is populated with packages
+
+        The default implementation of this method is to clear the cache and
+        re-save all packages, but the various subclasses may override this for
+        a more performant mechanism.
+
+        """
         self.clear_all()
         packages = self.storage.list(self.package_class)
         for pkg in packages:
--- a/pypicloud/cache/dynamo.py
+++ b/pypicloud/cache/dynamo.py
@@ -3,7 +3,7 @@ import logging
 import time
 from collections import namedtuple
 from datetime import datetime
-from itertools import imap
+from itertools import imap, chain
 from pkg_resources import parse_version
 
 from dynamo3 import ConditionalCheckFailedException, DynamoDBConnection
@@ -11,7 +11,7 @@ from pyramid.settings import asbool
 
 from .base import ICache
 from pypicloud.models import Package
-from pypicloud.util import EPOCH
+from pypicloud.util import EPOCH, retry
 
 
 try:
@@ -51,97 +51,55 @@ def _decide_between_versions(contender, current):
     return current
 
 
-def _clear_summary(summary):
-
-    """ Clear out the field of a PackageSummary object """
-    summary.stable = None
-    summary.unstable = '0'
-    summary.last_modified = EPOCH
-    return summary
-
-
-CacheUpdates = namedtuple(
-    'CacheUpdates',
+PackageUpdates = namedtuple(
+    'PackageUpdates',
     [
         'new_packages',
         'seen_packages',
+        'updated_packages',
         'stale_packages',
-        'new_summaries',
-        'seen_summaries',
-        'stale_summaries',
     ],
 )
 
 
-def calculate_cache_updates(
-    cache_packages,
-    cache_summaries,
-    storage_packages,
-    summary_factory,
+def calculate_package_updates(
+        cache_packages,
+        storage_packages,
 ):
-
     """ Calculates what changes need to happen in the cache to update it. """
 
     cached_pkg_by_filename = dict(
         (pkg.filename, pkg,)
         for pkg in cache_packages
     )
-    # Clear out non-key fields of the cached summaries to purge stable and
-    # unstable fields.
-    cached_summ_by_name = dict(
-        (summ.name, _clear_summary(summ),)
-        for summ in cache_summaries
-    )
 
     new_packages = set()
     seen_packages = set()
-    stale_packages = set()
-
-    new_summaries_by_name = dict()
-    seen_summaries = set()
+    updated_packages = set()
 
     for pkg in storage_packages:
-        try:
-            summ = cached_summ_by_name[pkg.name]
-        except KeyError:
-            summ = new_summaries_by_name.setdefault(
-                pkg.name,
-                summary_factory(pkg),
-            )
-        else:
-            seen_summaries.add(summ)
-
         try:
             cached_pkg = cached_pkg_by_filename[pkg.filename]
         except KeyError:
             new_packages.add(pkg)
-            summ.update_with(pkg)
         else:
-            if _decide_between_versions(
-                pkg,
-                cached_pkg,
-            ) is pkg:
-                cached_pkg.filename = pkg.filename
+            if _decide_between_versions(pkg, cached_pkg) is cached_pkg:
+                seen_packages.add(cached_pkg)
+            else:
                 cached_pkg.name = pkg.name
                 cached_pkg.version = pkg.version
                 cached_pkg.last_modified = pkg.last_modified
                 cached_pkg.data = dict(pkg.data)
 
-            seen_packages.add(cached_pkg)
-            summ.update_with(cached_pkg)
+                updated_packages.add(cached_pkg)
 
     stale_packages = set(cached_pkg_by_filename.itervalues()) - seen_packages
 
-    new_summaries = new_summaries_by_name.itervalues()
-    stale_summaries = set(cached_summ_by_name.itervalues()) - seen_summaries
-
-    return CacheUpdates(
+    return PackageUpdates(
         new_packages,
         seen_packages,
+        updated_packages,
         stale_packages,
-        new_summaries,
-        seen_summaries,
-        stale_summaries,
     )
 
 
@@ -233,34 +191,17 @@ class DynamoCache(ICache):
         return kwargs
 
     def reload_from_storage(self):
-        LOG.info("Recalculating cache.")
+        LOG.info("Recalculating package cache.")
 
-        cache_updates = calculate_cache_updates(
+        cache_updates = calculate_package_updates(
             self.engine.scan(DynamoPackage),
-            self.engine.scan(PackageSummary),
             self.storage.list(self.package_class),
-            PackageSummary,
         )
 
-        # HACK flywheel only checks for dirtyness on the first difference.
-        # If we do:
-        #     obj.a, obj.b = obj.b, obj.a
-        #     obj.a, obj.b = obj.b, obj.a
-        # We're going to get:
-        #     bool(obj.__dirty__) == True
-        def is_effectively_dirty(o):
-            """ Confirm that the fields marked on __dirty__ are true """
-            return any(
-                o.ddb_dump_cached_(f) != o.ddb_dump_field_(f)
-                for f in o.__dirty__ | set(o.__cache__.keys())
-            )
-        updated_packages = filter(is_effectively_dirty, cache_updates.seen_packages)
-        updated_summaries = filter(is_effectively_dirty, cache_updates.seen_summaries)
-
-        LOG.info("Cache recalculated. Persisting new version.")
+        LOG.info("Package cache recalculated. Persisting new version.")
 
         def do_or_skip(operation, coll, warn=True, **kwargs):
-            """ Do `operation`, ignore ConditionalCheckFailedExceptions """
+            """ Do `operation`, ignore ConditionalCheckFailedException """
             operator = getattr(self.engine, operation)
             for o in coll:
                 try:
@@ -274,50 +215,23 @@ class DynamoCache(ICache):
                         o,
                     )
 
-        def do_or_recalc_summaries(operation, coll, **kwargs):
-            """ Do `operation`, on ConditionalCheckFailedExceptions
-            recalculate the summary. """
-            operator = getattr(self.engine, operation)
-            for summ in coll:
-                try:
-                    operator(summ, **kwargs)
-                except ConditionalCheckFailedException:
-                    # If somebody else updated a summary, there's probably a
-                    # new package that got mutated after we read Dynamo. We
-                    # need to rebuild the whole summary.
-                    # Redoing it once will suffice in the overwhelming majority
-                    # of the cases. If not, it'll have to wait until the next
-                    # rebuild.
-
-                    LOG.warn(
-                        "Conflict while trying to %s %r. Recalculating.",
-                        operation,
-                        summ,
-                    )
-
-                    pkgs = self.all(summ.name)
-                    if not pkgs:
-                        do_or_skip('delete', [summ], raise_on_conflict=True)
-                        continue
-
-                    summ.refresh(consistent=True)
-                    summ = _clear_summary(summ)
-                    for pkg in pkgs:
-                        summ.update_with(pkg)
-
-                    do_or_skip('sync', [summ], raise_on_conflict=True)
-
-        do_or_skip('save', cache_updates.new_packages, overwrite=False, warn=False)
         # If somebody else updated a package, they're probably operating on
         # fresher data. In those cases, skip the change.
-        do_or_skip('sync', updated_packages, raise_on_conflict=True)
+        do_or_skip('save', cache_updates.new_packages, overwrite=False, warn=False)
+        do_or_skip('sync', cache_updates.updated_packages, raise_on_conflict=True)
         do_or_skip('delete', cache_updates.stale_packages, raise_on_conflict=True)
 
-        do_or_recalc_summaries('save', cache_updates.new_summaries, overwrite=False)
-        do_or_recalc_summaries('sync', updated_summaries, raise_on_conflict=True)
-        do_or_recalc_summaries('delete', cache_updates.stale_summaries, raise_on_conflict=True)
+        LOG.info("Package cache persisted. Updating summaries.")
 
-        LOG.info("Finished persisting cache.")
+        all_changed_packages = chain(
+            cache_updates.new_packages,
+            cache_updates.updated_packages,
+            cache_updates.stale_packages,
+        )
+        for name in set(p.name for p in all_changed_packages):
+            self._rebuild_summary_for_package_named(name)
+
+        LOG.info("Summaries updated. All done.")
 
     def fetch(self, filename):
         return self.engine.get(DynamoPackage, filename=filename)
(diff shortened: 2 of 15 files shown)


pkoch:
Now with most of the tests. Let me know what you think!

pypicloud/cache/dynamo.py
```diff
@@ -22,6 +24,82 @@
...
+            seen_summaries.add(summ)
+
+        summ.update_with(pkg)
+
+    new_summaries = new_summaries_by_name.itervalues()
+    stale_packages = set(cached_pkg_by_filename.itervalues()) - seen_packages
+    stale_summaries = set(cached_summ_by_name.itervalues()) - seen_summaries
+
+    return (
+        (
```
patricklucas: Use a namedtuple for this type?

pkoch: Yeah. I was trying to avoid creating a new data type just for one use site. However, this got pretty comical and you also felt the inclination.

Moving for that.

stevearc:
When I added the Dynamo caching backend, I put in a relatively simple implementation that worked in most cases but didn't pay too much attention to concurrency issues. I'm on board with making it more robust since it seems like several people are using it with multiple pypicloud servers. My main concern with this implementation is that it's a lot of code. Given Dynamo's consistency model, I'd feel a lot more confident about an implementation that uses two small atomic `add

For example, inside `do_or_recalc_summaries` you can still end up with an inconsistent PackageSummary table if the final `sync` operation fails and is skipped

pypicloud/cache/dynamo.py
```diff
@@ -111,20 +199,99 @@ def configure(cls, settings):
...
-        self.engine.delete_key(DynamoPackage, pkgs)
+        LOG.info("Recalculating cache.")
+
+        def normalize_last_mod(o):
+            """ Make it UTC without mized TZ comparisons """
+            # HACK Don't make flywheel compare mixed TZ-awareness objects.
+            d = o.last_modified.replace(tzinfo=UTC)
+            o.last_modified = None
+            o.last_modified = d
+            return o
```
stevearc: I think the correct solution here may be to either make everything else in pypicloud use a timezone-aware datetime, or to make flywheel play nice when changing a datetime to have/remove a timezone

pypicloud/cache/dynamo.py
```diff
@@ -22,6 +25,91 @@
...
+        try:
+            cached_pkg = cached_pkg_by_filename[pkg.filename]
+        except KeyError:
+            new_packages.add(pkg)
+        else:
+            seen_packages.add(cached_pkg)
+
+            # Skip over duplicate files (same filename, different paths).
+            # XXX Is this acceptable? How should I decide this?
+            if cached_pkg.last_modified > pkg.last_modified:
```
stevearc: I would be inclined to say that if the filename exists in the cache then the package can be ignored entirely. Do you have cases where the cache contains a filename but needs to update the other data?

stevearc:
As an example of what I mean, here's what the `save` operation might look like:

```
    def save(self, package):
        saved_package = False
        while True:
            summary = self.engine.get(PackageSummary, name=package.name)
            if summary is None:
                summary = PackageSummary(package)
            else:
                summary.update_with(package)

            if not saved_package:
                self.engine.save(package)
                saved_package = True
            try:
                self.engine.sync(summary)
            except CheckFailed:
                LOG.warning("Concurrent updates of %s summary", package.name)
            else:
                break
```
And then the 'reload_from_storage` would string together a series of these plus a similar delete operation. That still doesn't technically cover all of the potential concurrency issues, but given the access pattern from pypicloud it provides pretty good guarantees.

Any thoughts about that approach?

stevearc:
Been thinking about this some more. Another thing to consider is that the common race condition is new uploaded packages during the rebuild. If we assume a simple add/delete diff system and override the PackageSummary objects, the only data that could be out of sync is the 'latest stable/unstable version' and the 'last modified' time. That's a pretty unimportant feature in general and won't interfere with pip at all; the concurrently added packages will still show up and be installable.

pypicloud/cache/dynamo.py
pkoch: I do observe cases where people mistakenly upload packages to some key (they do the first upload to the wrong place, and then correct it, or they mistakenly upload a duplicate to some other path, and later delete that duplicate).

While this is not a major issue by any means, I'd like to be able to have defined as predictable behaviour from the system. Arguably, we should try to keep the number of different files for the same version to a minimum. Using the earliest/oldest file seems to be a scheme that's easy to both understand and implement.

pypicloud/cache/dynamo.py
pkoch: Yes, I agree, but we need both. Fixing this in flywheel avoids some code here (and is nice in general), but we still need to make sure we're doing the correct thing with datetimes. This should be done by just making every `datetime` TZ-aware at the cache and storage level. Moving for that.

pkoch:
That ends up being wasteful, as you're making the operation linear requests over you packages, and not batching at all.

I do agree that the current approach is the "full deal", and not necessarily the most desirable thing on the long run for this service. I need to take a better gander at the full system and deciding, but I think I'll converge on some variation of that.

From the little I could look around so far, it seems like there's not much point in being cache coherent on the summaries except for the full listing pages (be it in simple, API, or admin). So, I'd even be happy in dropping the concept altogether. If we can't drop that requirement, I'm certainly very open to make the summaries updates "non-sticklers". Let me build more context and get back to you on that.

What made you want to cache this in the first place?

stevearc:
Ha...I decided to cache that data a long time ago when it was strictly a SQL cache and S3 storage backend, before anything was made pluggable. At the time it seemed neat and relatively harmless. That's how it always is, right? -_-

The summary version and modified data is only used in the web display, I'm pretty sure. There's this tradeoff that I'm trying to navigate where the tighter consistency guarantees increases the likelihood of bugs due to the added complexity and adds to the maintenance burden by making the code harder to reason about. That's why I'm advocating for fewer guarantees and trying to find the ones that can be relaxed or ignored entirely.

Another way to decrease the risk would be to use a very simple reload algorithm by default and add a config option for 'strong_consistency' that does the more complex version.

pypicloud/cache/dynamo.py
pkoch: The flywheel side's PR is on mathcamp/flywheel#35.

pypicloud/cache/base.py
```diff
@@ -6,6 +6,7 @@
 from pyramid.settings import asbool
 
 import posixpath
+from flywheel.fields.types import UTC
```
stevearc: Since `flywheel` isn't a strict dependency of pypicloud (and I'd like to keep it that way), this import can fail on systems where only a `pip install pypicloud` has been run. I'd recommend just copying the UTC class (it was copied from [the python docs](https://docs.python.org/2/library/datetime.html) in the first place) and putting it into `util.py` or wherever.

pkoch: Ah, yes, makes perfect sense. Maybe we should be using pytz instead? Not a bad dep to have, and reduces such boilerplate. I'm open to both approaches, let me know which one you prefer.

pkoch:
That's exactly how it all starts. :)

I'm having good results from flat out removing `PackageSummary`. It looks mostly good, except it made `distinct` linear. From poking around in my UI, it seems to respond in a tolerable amount of time (~1s) with ~15k files in the cache. Not super thrilled about that, but not horrified (especially since pip and friends rarely list the whole thing).

We can make this faster if we make an index with only the name field and scan it. However, it seems like we would need to teach flywheel that trick (it very specifically doesn't want to scan indexes). I'm going to publish what I have right now, and we can always implement that if this becomes a problem in production.

Commits incoming!

pypicloud/cache/base.py
stevearc: Either is fine. I usually try to keep the dependencies to a minimum, but I realize that that's because all my other projects are libraries. For pypicloud it really doesn't matter.

stevearc:
The full table scan feels like the wrong thing to do here. Both the slow response time and high amount of consumed throughput make me think that it will make dynamo unsuited as a cache beyond a certain scale. I'd personally rather have it be fast for all operations, even if the displayed version numbers get out of sync after a `reload_from_storage`. I think the common use case will be frequent reads, common writes, and very rare reloads.

I don't think that a global index scan would be any faster, but I don't have benchmarks so if that is significantly faster it would be a possible solution. (flywheel was written way before global index scans were a thing, but it could be added if needed)

I like how this is looking; I think if you just add back in the bare minimum to get the `PackageSummary` to behave in most cases that'll be good.

pkoch:
Alright, I think I've achieved a decent compromise. The trick in this patch is basically to extract out the concept of rebuilding a summary and doing it right after the package updates (as opposed to intermingling), and to make it retry in case something changes the summary from under us.

I think this solution has lent itself to a nice reduction in complexity. I'm still a bit sad that I have to keep the `do_or_skip` to not write over other incoming updates but, short of synchronizing (with Dynamo's transaction or something external like Zoo Keeper, both of which I would consider overkill), I think we don't have much alternative here.

I'm getting to the bottom of my budget for this. I'm happy to accommodate some nip/tuck, but let me know  if this is not the direction you want to go so I can plan accordingly.

pypicloud/cache/dynamo.py
```diff
@@ -144,29 +255,33 @@ def summary(self):
...
-        summary = self.engine.get(PackageSummary, name=package.name)
-        if summary is not None and \
-            (summary.unstable == package.version or
-             summary.stable == package.version or
-             summary.last_modified == package.last_modified):
+    @retry(tries=3, exceptions=(ConditionalCheckFailedException,))
+    def _rebuild_summary_for_package_named(self, name):
+        """Rebuild the summary from the cached packages."""
+
+        pkgs = self.all(name)
```
stevearc: What do you think of doing strongly consistent reads here?

pypicloud/cache/dynamo.py
```diff
@@ -22,6 +26,86 @@
 LOG = logging.getLogger(__name__)
 
 
+def _decide_between_versions(contender, current, allow_overwrite):
```
stevearc: I don't think you need to respect `allow_overwrite` for this process. That setting is just there to prevent you from overwriting an existing package file, and I don't think it should matter at all if you're rebuilding the cache data.

pypicloud/cache/dynamo.py
```diff
@@ -22,6 +26,86 @@
...
+    elif contender.last_modified < current.last_modified:
+        if not allow_overwrite:
+            return contender
+    else:
+        def path_depth(package):
+            """Count the slashes in path, or return infinity on no path."""
+            path = package.data.get('path')
+            if not path:
+                return float('+inf')
+            return path.count('/')
```
stevearc: In the case where the cache & the storage both have an entry with the same filename, *but* the stored `path` is different, I think we always want to replace the `data` in the cache with the one from storage.

stevearc:
One question and two minor suggestions. Looks good!

pypicloud/cache/dynamo.py
pkoch: I think that's a very good call. Nice catch.

pypicloud/cache/dynamo.py
pkoch: I decided to also respect it during cache rebuilds for two reasons:
 * Some people might have setups where packages might just appear on S3
 * It confers some more determinism to solving the ordering conundrum.

Thinking about it again, since it doesn't provide full determinism (files a/pkg and b/pkg, if created at the same time, are still not orderable between them) and it introduced a conceptual dependency that's surprising for you, I'll just remove it and make the earliest win.

pypicloud/cache/dynamo.py
pkoch: That doesn't cover the case where you have two files with the same filename (but different paths) on storage. This methods basically encapsulates which of the two the rebuild prefers. The cache will have the path of the chosen file.

pypicloud/cache/dynamo.py
pkoch: Hmm. Dynamo doesn't like consistent queries on GSIs. Still making the summary get consistent (since we expect this to clash).

stevearc:
√
I'll re-look over the changes when I get a chance. Been a bit hosed recently, but I hope to get to it before the weekend.

pypicloud/cache/dynamo.py
```diff
@@ -22,6 +25,78 @@
...
+    if not path:
+        return float('+inf')
+    return path.count('/')
+
+
+def _decide_between_versions(contender, current):
+    """Decide between packages with the same filename on different paths."""
+
+    if contender.last_modified < current.last_modified:
+        return contender
```
stevearc: I think you need an `else` here because you only want to sort by path depth if the `last_modified` is equal

Also, maybe I'm going insane, but shouldn't this take the newest package not the oldest?

stevearc:
Just the one comment/question!

pypicloud/cache/dynamo.py
pkoch: D'oh! Yes, that got lost in the refactor. Also adding in a test to clarify that behaviour (since I should have been bitten by that in the tests, not in the code review).

This ordering is a good bit hard to reason for me too. Let's try to wrangle it!

Two files with the same filename but different path is clearly an operator mistake, and has to be dealt with in the least impactful and surprising manner. The two alternatives I see are:
* Using the latest, the operator will see their change take effect (and will not have their suspicion aroused). The users of the package are now being served a new file for the same exact version. This can lead to problems if the new package is not equivalent to the old one.
* Using the eldest, the operator will not see their change take effect, and will (ostensibly) start investigating why. The users of the package are not affected in any way.

If we compare the necessary work to bring reality to a desired state, we would have, respectively:
* All the engineering org that depends on that file is at risk. Someone has to notice there's a problem, the operator has to be notified of their mistake, remove the new package, let everyone affected know, and these people might have to follow up even further (if they published something that contains the package, like an AMI or something). Or, the files are equivalent and it's all good.
* The operator loses a few hours (probably less than that) investigating why their file upload didn't produce the desired change, and realize they need to bump the package instead (or take some other course of action). This happens even if the files are equivalent.

Given those two alternatives, I've went with using the eldest. There's less potential for fallout, and it makes the symptom happen close to its root cause. Considering the fallout might be unbound (and cascading), I think that justifies the price of having wierd goose chases (and the associated frustration) when the files are equivalent.

LMK if you disagree.

pkoch: Just realized that, for the fallout of the second case, if the files are equivalent, that won't send the operator on a goose chase. Since they're indiscernible, the operator will assume their actions got effected as expected. This means the expensive part of the consequences only happen when there's an actual mistake.

Now I'm even more set on that side! :P

pkoch: Just realized I can't make a failing test for this.

With a list of packages and a select element from that list, regardless of the the order of the list and the element we select, we're always going to converge on the earliest element. Even if the only ordering we have is partial (as is our case, we were only able to tell what's earlier, not equal).

However, not everybody has poset theory present when reading the code (neither did I, until I exhausted all the possible test formulations). Leaving the redundant check for clarity and to make it reusable in other contexts without gotchas.

pkoch:
Nothing terribly new. Here's the changed part: https://gist.github.com/pkoch/b3127c184ee55198a80a

pypicloud/cache/dynamo.py
stevearc: Ah, I was confused because I thought the goal was to get the freshest data. I'm totally behind you on making the behavior be the "least surprising". That seems very reasonable.

Thinking about it some more, if you're trying to make it the least surprising, wouldn't you just ignore files with the same filename that are in both cache & storage? I guess it would only cause the situation you mentioned if the paths were different. So I guess the gain from `_decide_between_versions` is that if you have the same filename in storage & cache, *but* the one in storage has a different `last_modified` (and the same `path`), then the `last_modified` in the cache will be updated.

I think it may be best to just take the one already in the cache if the filenames are equal. If you do want to keep this then I think you have to change it to first return the cached package if the paths are not equal, and then do any other comparisons after that.

pkoch: Re: same path: Derp, yeah. Your reasoning is well guided, but I actually would defend the other way around: if they have the same path, the storage always wins. In the "same path" specific case, we're not deciding between two distinct files, but rather two different versions in time of the same file. Even if we kept the old info around, it would be describing a file we can no longer serve. In this case, there's little pypicloud can do to prevent surprises, and overwriting a path should have been prevented through bucket policy. Good catch! :)

Re: decide based on cache: I like the stability of that approach a lot, but I see too many problems with it.

Considering the cache more than just a cache, since the values that it stores get decided based on both the storage and itself, is something that asks for trouble, I would say. Ostensibly, that makes updates noncommutative (since the initial state is always an empty cache), which will make this very sensitive to timing issues. Since we're working with eventually consistent things, I'm very phobic of that.

Off the top of my head, here are some bothersome complications:
 * An "absolute reset" (blowing away the dynamo tables and re-creating them) could yield different results than rebuilding it.
 * Racy scenarios yield different cache states with no obvious reason.
 * It's so much harder to reason about in general from the explosion in possibilities (since we now have to consider all possible timelines).

stevearc: Sorry, I don't mean to waste your time, but I am either misunderstanding some of your points or misreading your code. I'm going to try to nail down the parts I find confusing.

> if they have the same path, the storage always wins.

I don't think this is the case. That's what [the docstring on _decide_between_versions](https://github.com/mathcamp/pypicloud/pull/69/files#diff-cd8bb69a797363ac5abd920593f7114bR37) says, but it looks like it's being called on [every pair of packages](https://github.com/mathcamp/pypicloud/pull/69/files#diff-cd8bb69a797363ac5abd920593f7114bR86) where the same filename appears in both the cache and the storage, regardless of if the paths differ.

If the paths *do* differ, we want to keep the one that's currently in the cache to avoid nasty surprises. But in `_decide_between_versions` the priority is first decided by `last_modified` and then `path_depth`. Shouldn't the first operation be to compare paths and, if unequal, return `current`? (Preferably with a log warning that says that something fishy is going on)

> Considering the cache more than just a cache, since the values that it stores get decided based on both the storage and itself, is something that asks for trouble

Isn't the rebuild operation doing exactly that? It's comparing the versions of packages in the cache & storage before determining which updates get pushed to the cache.

pkoch: Not wasting time at all, we're figuring things out. The mere fact of having to put things down for another human has already convinced/dissuaded me of very important things!

I think I was unclear. What I meant is that you lead me to believe that was the right call, not what my patch actually does. Your assessment of what the code is currently doing is correct, and you are correct to suggest that we should compare pathness first and foremost.

To make sure we're on the same page, here's what I believe the code should look like:
```
def _decide_between_versions(cache, storage):
  if cache.filename != storage.filename:
    return current  # Trivial decision; pretty sure we agree on this.

  if cache.data['path'] == storage.data['path']:
    return contender  # Storage has new data for us on this path. Just accept it.

  # Here, we're dealing with two different files (since they have distinct paths)
  # with the same filename. Here are the two things I want to do:
  # * We should figure out a way to get them to decide on the same end state,
  #   regardless of who came first.
  # * We should always keep who comes first, in order to avoid surprises.
```

Ah! This is where I'm betraying myself. I want it both to be convergent and not-changing. This is incompatible (as I'm defining it right now). It's probably tolerable if it's a fraction of a millisecond (while two updates are trying to race each other), not so fine if it's during observable spans. That probably also explains why I confused you. I need some time to decide on a position to take here.

re what the cache is doing: not really (or it wasn't my intention). Cache rebuilds are supposed to make the cache reflect what we can observe in S3. Regardless of past state and events, given a snapshot of the bucket's contents, it should yield the same data on the cache. Sure, the mutation we do to it might be tailored to whatever is already in the cache, but the end state should always be the same (or converge to be the same, because eventual consistency and races).

This means that a "guided refresh" (by which I mean the informed approach in this patch) should yield the same results as an "absolute reset" (where we throw everything away and start with a fresh cache). That's what I was trying to advocate for.

pkoch: Wow, this is being particularly hard to keep all the details at once in my head. However, here's what I currently think is our best approach:

```
def _decide_between_versions(cache, storage):
  if cache.filename != storage.filename:
    return current  # Trivial decision; pretty sure we agree on this.

  if cache.data['path'] == storage.data['path']:
    return contender  # Storage has new data for us on this path. Just accept it.

  return sorted(
    [cache, storage],
    key=lambda o: (
      o.last_modified,
      o.data['path'],
    ),
  )[0]
```

I'm a bunch lost, and I might be leaving some case behind, but, I think this nails it all. The choice of using path (not using depth anymore) is pretty arbitrary. The only thing I need is a way to choose between different files that were uploaded at *exactly* the same second to S3. Since we've already asserted they're different, and strings are fully orderable, one of them has to always win the same way. The only way (I see) to change the file for a filename is to delete it. I can live with that.

Let me know if I stopped making sense somewhere or if I contradicted myself again. If we need to iterate this further, I vote we start writing down the properties we want observed as test cases. This way, we can have a machine tell us what we left behind.

stevearc: Okay great, I'm glad we're on the same page. Line-by-line:

```
if cache.filename != storage.filename:
```
Pretty sure that should never happen, as we're only comparing packages with the same filename. Is harmless, though.

```
if cache.data['path'] == storage.data['path']:
    return contender
```

√
(but use .get() because 'path' doesn't exist for all storage backends)

```
  return sorted(
    [cache, storage],
    key=lambda o: (
      o.last_modified,
      o.data['path'],
    ),
  )[0]
```

Not sure exactly the reasoning behind this. I think that there are two possible goals for the case where the paths are different:

1) **Least surprise to the package consumer**: In this case we should always `return current` because anything else will change the `path` stored in the cache, and pypicloud will begin serving a *different* file from S3. Preferably spit out a log line indicating what has happened.

2) **Same data as "absolute reset"**: In this case we should always `return contender` because anything else would leave different data in the cache & storage.

There is a separate but related issue that you pointed out, which is how to decide between multiple files from storage that have the same filename. Right now I don't think there's a code path that handles that at all. Since we're keeping `updated_packages` in a set, the set behavior is what's going to decide. The package `__eq__` method compares package name and version (though maybe it should be changed to compare filename) so I think the winner will be the first file returned by S3.

All of the caches currently have undefined behavior in this case, so I am okay with leaving it as such. If you want to make it deterministic then I'm fine with sorting on basically any field.

pkoch: Sorry for the belated answer.

I think we're actually defending the same thing. Let me write some tests to show that.

Meanwhile, @patricklucas also found a bug when rebuilding packages with the same package name and version, but different filenames (like wheels). Moving to fix that soon.

pkoch:
This is pretty old, and my interest in my interest in pypicloud is long gone (I don't work at Yelp anymore). Closing!

PR #71, opened 2016-03-07
Cloudfront support

pierrei:
To deliver package binaries faster I've built upon the S3 storage backend to support generating URLs to CloudFront instead of S3.

I've verified that this patch works and I'm curious about feedback to make it ready to merge into the official pypycloud distribution.

### DIFF
--- a/pypicloud/storage/s3.py
+++ b/pypicloud/storage/s3.py
@@ -8,6 +8,7 @@ from hashlib import md5
 from urllib import urlopen
 
 import boto.s3
+from boto.cloudfront import Distribution
 from boto.s3.key import Key
 from pyramid.httpexceptions import HTTPFound
 from pyramid.settings import asbool
@@ -84,15 +85,19 @@ class S3Storage(IStorage):
         kwargs['bucket'] = bucket
         return kwargs
 
+    def get_filename(self, package):
+        filename = package.name + '/' + package.filename
+        if self.prepend_hash:
+            m = md5()
+            m.update(package.filename)
+            prefix = m.digest().encode('hex')[:4]
+            filename = prefix + '/' + filename
+        return filename
+
     def get_path(self, package):
         """ Get the fully-qualified bucket path for a package """
         if 'path' not in package.data:
-            filename = package.name + '/' + package.filename
-            if self.prepend_hash:
-                m = md5()
-                m.update(package.filename)
-                prefix = m.digest().encode('hex')[:4]
-                filename = prefix + '/' + filename
+            filename = self.get_filename(package)
             package.data['path'] = self.bucket_prefix + filename
         return package.data['path']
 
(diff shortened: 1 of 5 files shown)


pypicloud/storage/s3.py
```diff
@@ -151,3 +157,47 @@ def open(self, package):
...
+            settings, 'storage.cloud_front_key_id', 'aws.cloud_front_key_id', '')
+
+        return kwargs
+
+    def get_url(self, package):
+        """ Get the fully-qualified CloudFront path for a package """
+        path = self.calculate_path(package)
+        url = self.cloud_front_domain + '/' + path
+
+        url = url.replace('+', '%2b')
```
stevearc: I'd prefer the more general `urllib.quote` here

setup.py
```diff
@@ -25,6 +25,7 @@
     'pyramid_duh>=0.1.1',
     'pyramid_jinja2',
     'pyramid_tm',
+    'rsa',
```
stevearc: I'm unfamiliar with this. Does the boto `create_signed_url` function depend on this module?

stevearc:
One suggestion and one question. Otherwise looks good!

Thanks for the PR! Cloudfront support is something that I've thought about for a while now (see #11) but never got around to implementing.

pypicloud/storage/s3.py
pierrei: The problem is that quote replaces stuff that shouldn't be replaced:

    >>> quote('https://domain.com/test/api/?param=val')
    'https%3A//domain.com/test/api/%3Fparam%3Dval'

I tried to use quote in the test setup that I have now and the links don't work anymore. I know for sure that the `+` sign needs to be quoted and there might be a few more characters that we need to replace as well, but replacing all of them will break the links.

setup.py
pierrei: A quote taken from [this question](http://stackoverflow.com/questions/2573919/creating-signed-urls-for-amazon-cloudfront#answer-16145598) on stack overflow:

boto now supports a dedicated create_signed_url method and the former binary dependency M2Crypto has recently been replaced with a [pure-Python RSA implementation](http://stuvel.eu/rsa) as well, see [Don't use M2Crypto for cloudfront URL signing](https://github.com/boto/boto/pull/1214).

So to give you an answer: yes, we do need the `rsa` dependency to make `create_signed_url` work :)

pierrei:
Thanks for the comments, I answered both of them. Also: is there any documentation that needs to be updated? Do you prefer to do that yourself or if not, could you point me to where updates need to be made?

setup.py
stevearc: :+1:

pypicloud/storage/s3.py
stevearc: Makes sense; I'm still leery of ad-hoc character escaping in the url. Where was the `+` appearing in the url? Is it part of the `cloud_front_domain`? Or the package filename? I don't think that `url` can have any query parameters at this point in the code. Would `urllib.quote`-ing all of those individually before composing with slashes have the desired behavior?

stevearc:
The docs that would need to be updated [are here](https://github.com/mathcamp/pypicloud/blob/876a722953f16382edff3b5a4dfc9675595deb2c/doc/topics/storage.rst). I don't mind updating myself it since it'll be a pretty small change, but feel free to take it on if you like!

stevearc:
Awesome! Merged. Thanks again for the PR. I'll get a new release out soon.

stevearc:
Released as [0.3.8](https://pypi.python.org/pypi/pypicloud/0.3.8)

pierrei:
Thanks for merging! Wonderful news! :smile:

PR #72, opened 2016-03-10
add S3 compatible API to storage

ahmet2mir:
Enable S3 compatible API (like ceph, cleversafe...) and update the doc, supporting secure layer and different calling format.


pypicloud/storage/s3.py
```diff
@@ -9,16 +9,24 @@
...
 import boto.s3
 from boto.s3.key import Key
+import boto.s3.connection
 from pyramid.httpexceptions import HTTPFound
 from pyramid.settings import asbool
 
 from .base import IStorage
 from pypicloud.models import Package
-from pypicloud.util import parse_filename, getdefaults
+from pypicloud.util import parse_filename, getdefaults, str2bool
```
stevearc: I think that the `asbool` function from `pyramid.settings` will do what you want

pypicloud/storage/s3.py
```diff
@@ -48,6 +56,17 @@ def configure(cls, settings):
                                  'aws.access_key', None)
         secret_key = getdefaults(settings, 'storage.secret_key',
                                  'aws.secret_key', None)
+        host = getdefaults(settings, 'storage.host',
+                           'aws.host', boto.s3.connection.NoHostProvided)
+        is_secure = getdefaults(settings, 'storage.is_secure',
+                                'aws.is_secure', True)
+        calling_format = getdefaults(settings, 'storage.calling_format',
+                                     'aws.calling_format',
+                                     'SubdomainCallingFormat')
```
stevearc: You can just use the standard `settings.get('storage.calling_format', 'SubdomainCallingFormat')` syntax for these options. The only reason the others are using `getdefaults()` is to maintain backwards compatibility with old config options. You'll notice that `aws.access_key` and the others aren't even listed in the documentation anymore.

benh57:
Was just about to implement this myself -- thanks for doing it! Would love to see this feature in master

ahmet2mir:
@stevearc, I fixed as you suggest :)

@benh57 you're welcome :)

stevearc:
Merged and released in [0.3.9](https://pypi.python.org/pypi/pypicloud/0.3.9)

PR #77, opened 2016-03-24
Fix missing newline in config template

patricklucas:
Fix missing newline in config template

Before:
```ini
[uwsgi]
...
enable-threads = truesocket = 127.0.0.1:3031
```

After:
```ini
[uwsgi]
...
enable-threads = true
socket = 127.0.0.1:3031
```


stevearc:
Thanks!

PR #84, opened 2016-05-03
adds __pypicloud_ldap_admin__ as an admin service account

a-tal:
since the service account is configured via full DN, this adds the user name `__pypicloud_ldap_admin__` which is tied to the service account DN. if you know the password to pypicloud's ldap service account you can probably get admin access via other methods anyway... handy to have a service account though.

if you think it should be a different name I'm not very particular to `__pypicloud_ldap_admin__`, just wanted it to be something unlikely to be a regular account name

### DIFF
--- a/doc/topics/access_control.rst
+++ b/doc/topics/access_control.rst
@@ -356,3 +356,9 @@ will depend on your LDAP setup.
 
 A list of DNs to search for who should be considered an admin. It uses the
 ``admin_field`` to determine the source of admin DNs in the returned record(s).
+
+``auth.ldap.service_account``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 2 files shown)


stevearc:
Is the purpose of this so that anyone who knows the password for the LDAP service account can use that to log in as a pypicloud admin? If so I'd suggest putting it behind a setting so it can be tweaked if anyone doesn't want that behavior. A setting has the side benefit of allowing you to *completely ignore* the naming problem for the magic user.

a-tal:
yeah, that's basically the use case. fair point on making it configurable...

stevearc:
Nice. Thanks!

stevearc:
Released in version [0.3.12](https://pypi.python.org/pypi/pypicloud/0.3.12)

PR #87, opened 2016-05-11
Properly normalize names according to pep503

### LINKED ISSUE #86: Normalize package names
In pip 8.1.1 this works:

```
$ pip --version
pip 8.1.1 from /tmp/venv2/lib/python3.5/site-packages (python 3.5)
$ pip install repoze.lru -i https://pypi.yelpcorp.com/simple
... 
$ echo $?
0
```

In pip 8.1.2 it normalizes names:

```
$ pip --version
pip 8.1.2 from /tmp/venv2/lib/python3.5/site-packages (python 3.5)
$ pip install repoze.lru -i https://pypi.yelpcorp.com/simple -v
Collecting repoze.lru
  1 location(s) to search for versions of repoze.lru:
  * https://pypi.yelpcorp.com/simple/repoze-lru/
  Getting page https://pypi.yelpcorp.com/simple/repoze-lru/
  Looking up "https://pypi.yelpcorp.com/simple/repoze-lru/" in the cache
  No cache entry available
  Starting new HTTPS connection (1): pypi.yelpcorp.com
  "GET /simple/repoze-lru/ HTTP/1.1" 404 137
  Status code 404 not in [200, 203, 300, 301]
  Could not fetch URL https://pypi.yelpcorp.com/simple/repoze-lru/: 404 Client Error: Not Found for url: https://pypi.yelpcorp.com/simple/repoze-lru/ - skipping
  Could not find a version that satisfies the requirement repoze.lru (from versions: )
```

asottile:
Resolves #86

### DIFF
--- a/pypicloud/util.py
+++ b/pypicloud/util.py
@@ -1,5 +1,6 @@
 """ Utilities """
 import posixpath
+import re
 
 import logging
 from distlib.locators import Locator, SimpleScrapingLocator
@@ -32,7 +33,9 @@ def parse_filename(filename, name=None):
 
 def normalize_name(name):
     """ Normalize a python package name """
-    return name.lower().replace('-', '_')
+    # Lifted directly from PEP503:
+    # https://www.python.org/dev/peps/pep-0503/#id4
+    return re.sub(r"[-_.]+", "-", name).lower()
 
 
 class BetterScrapingLocator(SimpleScrapingLocator):
--- a/tests/test_util.py
+++ b/tests/test_util.py
@@ -56,3 +56,12 @@ class TestScrapers(unittest.TestCase):
         locator.prefer_wheel = False
         self.assertTrue(locator.score_url('http://localhost/mypkg-1.1.whl') <
                         locator.score_url('http://localhost/mypkg-1.1.tar.gz'))
+
+
+class TestNormalizeName(unittest.TestCase):
+
+    """ Tests for normalize_name """
+
+    def test_normalize_namespace_package(self):
+        """ Namespace packages must be normalized according to PEP503 """
+        self.assertEqual(util.normalize_name('repoze.lru'), 'repoze-lru')


stevearc:
Thanks! I'll need to think about this a bit before merging. Have to make sure it's backwards-compatible with anyone that already has stored packages or that there's a clear and easy migration path.

mike-hart:
Our packages were affected by the fallout from pip 8.1.1-8.1.2 (we have package names which include ".") so I'm eagerly awaiting this too :) @stevearc - no pressure, just, you know...

If existing packages are negatively affected by this change (entries already present in S3), is it difficult to re-upload/re-process them?

SideshowAlex:
@stevearc - I can confirm that this patch is not enough for users with existing packages. We had success with reuploading packages with this patch is in place.

SideshowAlex:
This patch also doesn't cover the case where packages are uploaded via the UI. The upload function in views/simple.py needs to normalize the name in for the web UI

```python
--- a/pypicloud/views/simple.py
+++ b/pypicloud/views/simple.py
@@ -24,6 +24,7 @@ def upload(request, content, name=None, version=None):
     # Direct uploads from the web UI go here, and don't have a name/version
     if name is None or version is None:
         name, version = parse_filename(content.filename)
+        name = normalize_name(name)
     else:
         name = normalize_name(name)
     if action == 'file_upload':
```

I can submit a pull request in the the next week, if it isn't fixed as part of this patch

stevearc:
Thanks @SideshowAlex, I'll incorporate that when I get a full solution in.

There's an additional complication that I have to resolve. When using `pypi.fallback = cache`, downloading upstream packages with legacy names fails because distlib doesn't do the PEP 503 normalization either. I've submitted a PR to distlib here: https://bitbucket.org/vinay.sajip/distlib/pull-requests/7/update-name-comparison-to-match-pep-503/diff

Here's my plan:
* [ ] Make this change backwards-compatible or add a migration path
* [ ] Grab the above patch from @SideshowAlex 
* [ ] Test everything thoroughly
* [ ] Either update distlib required version or add a monkey-patch with the fix

I'm hoping to get this all completed by the end of the weekend.

stevearc:
Done and released as version [0.4.0](https://pypi.python.org/pypi/pypicloud/0.4.0)

PR #92, opened 2016-05-27
Don't allow anonymous ldap binding

scizo:
At least against ActiveDirectory I am able to use basic auth
with an empty password and a valid username and python-ldap
does an anonymous bind to the server and does not throw an
ldap.INVALID_CREDENTIALS error. This allows anyone that
knows a valid username to use the server with basic auth. This
simple guard prevents this from happening.


stevearc:
Good catch, thanks.

PR #94, opened 2016-07-11
Add autofocus to search field

chriskuehl:
This makes the browser autofocus on the search field when loading the page.


stevearc:
Thanks!

PR #107, opened 2017-01-08
Pip search

### LINKED ISSUE #104: Server doesn't return expected format for search
Hey all,

Currently getting these errors when doing search against our own server (I also did them against the demo server for proof and to post here):
```
[~]$ pip search -i https://pypi.stevearc.com/pypi/ requests
HTTP error 400 while getting https://pypi.stevearc.com/pypi/
Exception:
Traceback (most recent call last):
  File "/opt/clearsky/lib/python2.7/site-packages/pip/basecommand.py", line 215, in main
    status = self.run(options, args)
  File "/opt/clearsky/lib/python2.7/site-packages/pip/commands/search.py", line 45, in run
    pypi_hits = self.search(query, options)
  File "/opt/clearsky/lib/python2.7/site-packages/pip/commands/search.py", line 62, in search
    hits = pypi.search({'name': query, 'summary': query}, 'or')
  File "/opt/clearsky/lib/python2.7/xmlrpclib.py", line 1240, in __call__
    return self.__send(self.__name, args)
  File "/opt/clearsky/lib/python2.7/xmlrpclib.py", line 1599, in __request
    verbose=self.__verbose
  File "/opt/clearsky/lib/python2.7/site-packages/pip/download.py", line 776, in request
    response.raise_for_status()
  File "/opt/clearsky/lib/python2.7/site-packages/pip/_vendor/requests/models.py", line 862, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
HTTPError: 400 Client Error: Bad Request for url: https://pypi.stevearc.com/pypi/
```
It looks to me as though PyPICloud does not have the correct file expected by PIP when it does a POST to search. Has anyone else run into this issue? If so, I'd like to formulate a fix. This is going to be absolutely necessary for us to consider deploying this at work, and it seems to be the only downside we've found so far.

MichaelPalmer1:
Fixes #104.

I took a quick stab at getting pip search working. I've tested this in Python 2.7 using SQLite, Redis, S3, DynamoDB, and file system with no issues.

Since pip search returns a summary of the package, as well as its name and version, I added a summary field to the cache so it could be searched. For local filesystem, I have it storing this in a metadata JSON file so the cache can be repopulated. Since S3 supports metadata attributes, I just set it as a key there instead of creating a file.

Let me know if there are any areas for improvement here.

### DIFF
--- a/pypicloud/cache/base.py
+++ b/pypicloud/cache/base.py
@@ -224,6 +224,10 @@ class ICache(object):
 
                 # Search package summaries
                 for query in summary_queries:
+                    # Skip if there is not a package summary
+                    if not package.summary:
+                        continue
+
                     # Look for this query anywhere in the package summary
                     if query.lower() in package.summary.lower():
                         # Found a match, adding to the packages_found
--- a/pypicloud/cache/base.py
+++ b/pypicloud/cache/base.py
@@ -193,7 +193,49 @@ class ICache(object):
             Type of query to perform. By default, pip sends "or".
 
         """
-        raise NotImplementedError
+        name_queries = criteria.get('name', [])
+        summary_queries = criteria.get('summary', [])
+        packages = []
+        packages_found = set()
+
+        for key in self.distinct():
+            # Search all versions of this package key
+            for package in self.all(key):
+                # Skip this result if it has already been selected
+                if package.name in packages_found:
+                    continue
+
+                # Search package names
+                for query in name_queries:
+                    # Look for this query anywhere in the package name
+                    if query.lower() in package.name.lower():
+                        # Found a match, adding to the packages_found
+                        # set and generating a result
+                        packages_found.add(package.name)
+                        packages.append({
+                            'name': package.name,
+                            'summary': package.summary,
+                            'version': package.version,
+                        })
+
+                # Skip this result if it was selected by the name search
+                if package.name in packages_found:
+                    continue
+
+                # Search package summaries
+                for query in summary_queries:
+                    # Look for this query anywhere in the package summary
+                    if query.lower() in package.summary.lower():
+                        # Found a match, adding to the packages_found
+                        # set and generating a result
+                        packages_found.add(package.name)
+                        packages.append({
+                            'name': package.name,
+                            'summary': package.summary,
+                            'version': package.version,
+                        })
+
+        return packages
 
     def summary(self):
         """
--- a/pypicloud/cache/base.py
+++ b/pypicloud/cache/base.py
@@ -8,8 +8,7 @@ from pyramid.settings import asbool
 import posixpath
 from pypicloud.models import Package
 from pypicloud.storage import get_storage_impl
-from pypicloud.util import parse_filename, normalize_name
-
+from pypicloud.util import create_matcher, parse_filename, normalize_name
 
 LOG = logging.getLogger(__name__)
 
@@ -198,6 +197,10 @@ class ICache(object):
         packages = []
         packages_found = set()
 
+        # Create matchers for the queries
+        match_name = create_matcher(name_queries, query_type)
+        match_summary = create_matcher(summary_queries, query_type)
+
         for key in self.distinct():
             # Search all versions of this package key
             for package in self.all(key):
@@ -206,38 +209,34 @@ class ICache(object):
                     continue
 
                 # Search package names
-                for query in name_queries:
-                    # Look for this query anywhere in the package name
-                    if query.lower() in package.name.lower():
-                        # Found a match, adding to the packages_found
-                        # set and generating a result
-                        packages_found.add(package.name)
-                        packages.append({
-                            'name': package.name,
-                            'summary': package.summary,
-                            'version': package.version,
-                        })
+                if match_name(package.name):
+                    # Found a match, adding to the packages_found
+                    # set and generating a result
+                    packages_found.add(package.name)
+                    packages.append({
+                        'name': package.name,
+                        'summary': package.summary,
+                        'version': package.version,
+                    })
 
                 # Skip this result if it was selected by the name search
                 if package.name in packages_found:
                     continue
 
-                # Search package summaries
-                for query in summary_queries:
-                    # Skip if there is not a package summary
-                    if not package.summary:
-                        continue
+                # Skip if there is not a package summary
+                if not package.summary:
+                    continue
 
-                    # Look for this query anywhere in the package summary
-                    if query.lower() in package.summary.lower():
-                        # Found a match, adding to the packages_found
-                        # set and generating a result
-                        packages_found.add(package.name)
-                        packages.append({
-                            'name': package.name,
-                            'summary': package.summary,
-                            'version': package.version,
-                        })
+                # Search package summaries
+                if match_summary(package.summary):
+                    # Found a match, adding to the packages_found
+                    # set and generating a result
+                    packages_found.add(package.name)
+                    packages.append({
+                        'name': package.name,
+                        'summary': package.summary,
+                        'version': package.version,
+                    })
 
         return packages
 
--- a/pypicloud/cache/base.py
+++ b/pypicloud/cache/base.py
@@ -72,7 +72,7 @@ class ICache(object):
         for pkg in packages:
             self.save(pkg)
 
-    def upload(self, filename, data, name=None, version=None):
+    def upload(self, filename, data, name=None, version=None, summary=None):
         """
         Save this package to the storage mechanism and to the cache
 
@@ -88,6 +88,8 @@ class ICache(object):
         version : str, optional
             The version number of the package (if not provided, will be parsed
             from filename)
+        summary : str, optional
+            The summary of the package
 
         Returns
         -------
@@ -107,7 +109,7 @@ class ICache(object):
         old_pkg = self.fetch(filename)
         if old_pkg is not None and not self.allow_overwrite:
             raise ValueError("Package '%s' already exists!" % filename)
-        new_pkg = self.package_class(name, version, filename)
+        new_pkg = self.package_class(name, version, filename, summary=summary)
         self.storage.upload(new_pkg, data)
         self.save(new_pkg)
         return new_pkg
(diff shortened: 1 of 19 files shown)


stevearc:
Hey, sorry for the long silence; I have had a very bad couple of months. I'm coming out of it now and I'm going to try to be more responsive!

First off: thanks for the PR! This is awesome! There are some minor changes that I'd like to see, and I'll go through and leave those comments inline.

pypicloud/cache/base.py
```diff
@@ -170,6 +172,71 @@ def distinct(self):
...
+
+        query_type : str
+            Type of query to perform. By default, pip sends "or".
+
+        """
+        name_queries = criteria.get('name', [])
+        summary_queries = criteria.get('summary', [])
+        packages = []
+        packages_found = set()
+
```
stevearc: This method doesn't respect the `query_type`. To support that, I'd suggest a slightly different mechanism for doing the matching. Something like:
```python
def create_matcher(queries, query_type):
    queries = [q.lower() for q in queries]
    if query_type == 'or':
        return lambda x: any((q in x.lower() for q in queries))
    else:
        return lambda x: all((q in x.lower() for q in queries))

match_name = create_matcher(name_queries, query_type)
match_summary = create_matcher(summary_queries, query_type)
```

And then below you can just use `if match_name(package.name): ...`

pypicloud/cache/base.py
```diff
@@ -170,6 +172,71 @@ def distinct(self):
...
+                    # Look for this query anywhere in the package name
+                    if query.lower() in package.name.lower():
+                        # Found a match, adding to the packages_found
+                        # set and generating a result
+                        packages_found.add(package.name)
+                        packages.append({
+                            'name': package.name,
+                            'summary': package.summary,
+                            'version': package.version,
+                        })
```
stevearc: I think that here it makes more sense to simply append the `package` directly. I think the only place the return value is used right now is in the search endpoint, and if you return the python objects directly they will be serialized with the `__json__()` method. That will add the additional `last_modified` and `url` fields, but if those don't cause problems for pip then that might be a cleaner way to do it.

pypicloud/cache/base.py
```diff
@@ -170,6 +172,71 @@ def distinct(self):
...
+
+                # Skip this result if it was selected by the name search
+                if package.name in packages_found:
+                    continue
+
+                # Search package summaries
+                for query in summary_queries:
+                    # Look for this query anywhere in the package summary
+                    if query.lower() in package.summary.lower():
```
stevearc: I think we have to handle the case when the summary is `None`

pypicloud/storage/files.py
```diff
@@ -26,20 +29,42 @@ def configure(cls, settings):
...
         return kwargs
 
-    def get_path(self, package):
-        """ Get the fully-qualified file path for a package """
+    def get_path(self, package, metadata=False):
+        """
+        Get the fully-qualified file path for a package and its metadata file.
+        """
+        if metadata:
+            filename = self.METADATA_FILE
```
stevearc: This is going to use the same metadata file for all packages of the same name and version. I think in most cases it'll be fine, but it feels weird that the behavior will be different from S3. Also there are some edge cases, such as:
1) Upload package-1.0.tar.gz
2) Upload package-1.0.whl
3) Delete package-1.0.whl

Now there is no metadata file for the remaining package

Changing this will also require changes in the `list()` method too, though

pypicloud/storage/files.py
```diff
@@ -26,20 +29,42 @@ def configure(cls, settings):
...
 
     def list(self, factory=Package):
         for root, _, files in os.walk(self.directory):
+            metadata = {}
+
+            # Read the metadata file
+            if self.METADATA_FILE in files:
+                with open(os.path.join(root, self.METADATA_FILE), 'r') as mfile:
+                    try:
+                        metadata = json.loads(mfile.read())
```
stevearc: Can just do `metadata = json.load(mfile)`

pypicloud/storage/files.py
```diff
@@ -48,12 +73,24 @@ def download_response(self, package):
...
         uid = os.urandom(4).encode('hex')
-        tempfile = os.path.join(destdir, '.' + package.filename + '.' + uid)
+
+        # Store metadata as JSON. This could be expanded in the future
+        # to store additional metadata about a package (i.e. author)
+        tempfile = os.path.join(destdir, '.metadata.' + uid)
+        metadata = {'summary': package.summary}
+        with open(tempfile, 'w') as mfile:
+            json_data = json.dumps(metadata)
+            mfile.write(json_data)
```
stevearc: `json.dump(metadata, mfile)`

pypicloud/storage/files.py
```diff
@@ -62,7 +99,9 @@ def upload(self, package, data):
 
     def delete(self, package):
         filename = self.get_path(package)
+        meta_file = self.get_path(package, metadata=True)
         os.unlink(filename)
+        os.unlink(meta_file)
```
stevearc: This will throw an `OSError` if the metadata file is not present

pypicloud/views/simple.py
```diff
@@ -31,13 +32,23 @@ def upload(request, content, name=None, version=None):
...
 
 
+@xmlrpc_method(endpoint='pypi')
+def search(request, criteria, query_type):
+    """
+    Perform searches from pip. This handles XML RPC requests to the "pypi"
+    endpoint (configured as /pypi/) that specify the method "search".
+
+    """
+    return request.db.search(criteria, query_type)
```
stevearc: It's important that the search only returns packages that the user has read access to. You can see several examples of how to do this [in this file](https://github.com/stevearc/pypicloud/blob/8138668402e58e5ecde25369040711de26baed73/pypicloud/views/simple.py#L50). One particular edge case to pay attention to is that with HTTP, pip will only send up the username/password if the server requests it by returning a 403. This means that in a lot of places we check if the user is logged in, and if not we [request the credentials](https://github.com/stevearc/pypicloud/blob/8138668402e58e5ecde25369040711de26baed73/pypicloud/views/simple.py#L156). I don't know what the behavior is with the RPC endpoint, but it's worth checking.

tests/test_cache.py
```diff
@@ -140,6 +140,8 @@ def test_abstract_methods(self):
             cache.clear_all()
         with self.assertRaises(NotImplementedError):
             cache.save(make_package())
+        with self.assertRaises(NotImplementedError):
+            cache.search({'name': [], 'summary': []}, 'or')
```
stevearc: I actually don't understand how this test is passing. Isn't there an implemented `search()` method on the ICache class?

tests/test_cache.py
```diff
@@ -283,6 +285,40 @@ def test_distinct(self):
...
+            {
+                'name': 'mypkg2',
+                'version': '1.3.4',
+                'summary': 'summary'
+            }
+        ]
+        packages = self.db.search(criteria, 'or')
+        packages.sort(key=lambda item: (item['name'], item['version']))
+        expected.sort(key=lambda item: (item['name'], item['version']))
+        self.assertListEqual(packages, expected)
```
stevearc: I think you may be able to use `self.assertItemsEqual` and skip the sorting

stevearc:
I'm going to go ahead and merge this into the 0.4 branch and make the changes I suggested myself so I can get a release out soon. Let me know if you have any concerns!

MichaelPalmer1:
Sounds good! Sorry I haven't been able to make any of the changes. I've been a little busy too lately. If you have any questions, let me know and I'll try to get back to you quickly.

pypicloud/views/simple.py
```diff
@@ -31,13 +32,23 @@ def upload(request, content, name=None, version=None):
...
-                                     version=version)
+                                     version=version, summary=summary)
         except ValueError as e:
             return HTTPBadRequest(*e.args)
     else:
         return HTTPBadRequest("Unknown action '%s'" % action)
 
 
+@xmlrpc_method(endpoint='pypi')
+def search(request, criteria, query_type):
```
MichaelPalmer1: I also wanted to throw in a note about this - I have separately been working on configuring the search to also search the fallback url (`pypi.fallback` setting) and return those results so it is not limited to just packages uploaded to the local pypi server. I think I have it working fairly well, but I'll do a separate PR for it.

MichaelPalmer1:
Hey @stevearc, I went ahead and fixed a couple of the things you mentioned.

pypicloud/storage/files.py
MichaelPalmer1: Resolved

pypicloud/storage/files.py
MichaelPalmer1: Resolved

pypicloud/storage/files.py
MichaelPalmer1: Resolved

pypicloud/cache/sql.py
```diff
@@ -163,6 +164,62 @@ def distinct(self):
...
+                condition = field.like('%' + query + '%')
+                column_conditions.append(condition)
+
+            # Check if conditions for this column were generated
+            if column_conditions:
+                if query_type == 'and':
+                    operator = and_
+                else:
+                    operator = or_
+                conditions.append(operator(*column_conditions))
```
MichaelPalmer1: I reworked the logic here to match the logic that the other cache backends used. I explained it in the docstring.

Let me know if this makes sense to you.

pypicloud/cache/base.py
MichaelPalmer1: Resolved

pypicloud/cache/base.py
MichaelPalmer1: Resolved

tests/test_cache.py
MichaelPalmer1: It turns out that the exception is being raised by `self.distinct()` on line 201 in cache/base.py: https://github.com/stevearc/pypicloud/pull/107/files/a70236fc78730ca1e1091a9a7b18f395bdda8ac0#diff-4928d680f66969deaa7e1b09ab1c558cR201

This probably is not a good test, so I will take it out.

setup.py
```diff
@@ -22,10 +22,11 @@
     'paste',
     'passlib',
     'pycrypto',
-    'pyramid',
+    'pyramid~=1.7.5',
```
MichaelPalmer1: Pyramid 1.8+ and webtest 2.0.24+ appear to have dropped support for Python 2.6, so I pinned their versions.

stevearc:
Thanks for the update! I'm going to merge this in to a new 0.5 branch on the commandline, patch some things on top of it, and then do a release. Might take a few days, but I'm starting now.

Just so you know what's going on: this will have to be released at least with a minor version bump since the SQL cache isn't backwards compatible and will need to be regenerated. I've also been working on a port to boto3 so we can support python 3, and I'm hoping to ship that out with a major version bump after the search stuff goes out.

stevearc:
Released in version [0.5.0](https://pypi.python.org/pypi/pypicloud/0.5.0)

Issue #124, opened 2017-05-16
Package files with escaped HTML characters don't load properly


triplepoint:
Given a package file named:
`something.something.ops-2016.9.9.23.2.47%2Bbee0714.tar.gz`

With a package METADATA `version`:
`2016.9.9.23.2.47+bee0714`

And having published the package to pypicloud with `twine`:
`twine upload -r my_repo_definition --skip-existing something.something.ops-2016.9.9.23.2.47%2Bbee0714.tar.gz`

I can see the file in the pypi UI, with the above file name and version.

Attempting to install the package with pip:
```
virtualenv env
source env/bin/activate
pip install -i https://my_pypicloud_server/simple/ -vvvv "something.something.ops==2016.6.15.5.33.2+21c64f5.dirty"
```

yields (replacing a few sensitive strings with placeholders):
```
pip install "something.something.ops==2016.6.15.5.33.2+21c64f5.dirty" -i https://my-pypi-cloud-redacted.com/pypi/ -vvvv
Collecting something.something.ops==2016.6.15.5.33.2+21c64f5.dirty
  1 location(s) to search for versions of something.something.ops:
  * https://my-pypi-cloud-redacted.com/pypi/something-something-ops/
  Getting page https://my-pypi-cloud-redacted.com/pypi/something-something-ops/
  Looking up "https://my-pypi-cloud-redacted.com/pypi/something-something-ops/" in the cache
  No cache entry available
  Starting new HTTPS connection (1): my-pypi-cloud-redacted.com
  "GET /pypi/something-something-ops/ HTTP/1.1" 200 577
  Updating cache with response from "https://my-pypi-cloud-redacted.com/pypi/something-something-ops/"
  Analyzing links from page https://my-pypi-cloud-redacted.com/pypi/something-something-ops/
    Found link https://my-pypi-cloud-redacted.com/api/package/something-something-ops/something.something.ops-2016.1.22.20.3.13%2B124b629.dirty.tar.gz (from https://my-pypi-cloud-redacted.com/pypi/something-something-ops/), version: 2016.1.22.20.3.13+124b629.dirty
    Found link https://my-pypi-cloud-redacted.com/api/package/something-something-ops/something.something.ops-2016.1.23.3.0.0%2B190054c.dirty.tar.gz (from https://my-pypi-cloud-redacted.com/pypi/something-something-ops/), version: 2016.1.23.3.0.0+190054c.dirty
[...]

PR #127, opened 2017-05-24
Undo the URL decoding that's happening in pyramid requests, in order …

triplepoint:
…to recover the actual file name of packages in the API endpoints.

Addresses issue #124 , but it's not clear to me that this is the best solution.  It's at least _a_ solution though, so I figure this can get the conversation going towards the best fix.

### DIFF
--- a/pypicloud/views/api.py
+++ b/pypicloud/views/api.py
@@ -12,6 +12,7 @@ from pyramid.security import NO_PERMISSION_REQUIRED
 from pyramid.view import view_config
 from pyramid_duh import argify, addslash
 from six.moves.urllib.request import urlopen  # pylint: disable=F0401,E0611
+from six.moves.urllib.parse import quote_plus
 
 from .login import handle_register_request
 from pypicloud.route import (APIResource, APIPackageResource,
@@ -69,7 +70,7 @@ def fetch_dist(request, package_name, package_url):
              permission='read')
 def download_package(context, request):
     """ Download package, or redirect to the download link """
-    package = request.db.fetch(context.filename)
+    package = request.db.fetch(quote_plus(context.filename))
     if not package:
         if request.registry.fallback != 'cache':
             return HTTPNotFound()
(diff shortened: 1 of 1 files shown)


stevearc:
Thanks for doing all the detective work to find out where the escaping was going wrong! I think this is an okay solution, but it seems like the root problem is that the filename is escaped before being stored in the cache. I would expect it to be stored unescaped, and then escaped on demand when displayed.

I don't have the time right now to dig into that, but I can probably do it this coming weekend. If it turns out to be a fruitless pursuit I'd be fine with merging your solution.

I'll also have to put some time into figuring out why the tests are failing because it's clearly unrelated. Probably due to a new release of boto or moto or something.

triplepoint:
Yea, i know what you mean.  I think the start of the problem is that `setuptools` url-encodes the filename when it generates it, when you first do `setup.py sdist`.  So the pypicloud cache is really just caching the file with the name it was given.  That S3 goes on to re-encode it is just due to it not realizing that the `%` is part of an escape code in an already-escaped filename.  That all more or less works, though, because `boto`'s url-generation code allows for that situation and double-encodes the generated URL as well.

The problem here is when pyramid gets a url-encoded file name request from pip, and tries to help by decoding the file name.  That "unwinds" one layer of the encoding, breaking a bunch of assumptions up the line.

stevearc:
Strangely enough, it all appears to work fine for me. Using postgres & S3, if I `twine upload dist/mypackage-0.1+aaaabbbb.tar.gz` and then `pip install mypackage==0.1+aaaabbbb` it installs no problem. Looking inside the DB shows that the filename and version are not url-encoded. I tried uploading with twine and with `setup.py sdist upload`. I'm not sure what would cause this behavior to be different between our setups. Maybe I missed something with the repro steps? Perhaps we have different versions of pyramid or its dependent libraries? I'm running
```
pyramid==1.8.3
WebOb==1.7.2
PasteDeploy==1.5.2
```

triplepoint:
Oooph, yeah, this may've been the result of a bad internal script on our end that inserted a url-encoding step when it was collecting all the packages to export from our old pip server and import into our new pypicloud server.  

I'm reimporting all quarter million or so packages, and will test with the stock version of the code to see if that was indeed the problem.

I'll post back here and close this PR if testing proves out that theory.  Sorry for the wild goose chase.

triplepoint:
Ok, I fixed our export/import script that had the unintended URL-encode in it (that'll teach me to scrape HTML pages for links without taking a close look).

Turns out that was the culprit.  Write a quick script to crawl the hashed S3 directories and remove url-encoded packages.  Rebuilt the Postgres cache database and tested with a few package files with `+` in their name, and no problems.  Looks like this one was entirely on our end.

Sorry again for the runaround!

Issue #126, opened 2017-05-24
SQL cache `version` field too small for some packages


triplepoint:
Wheels have started showing up in pypi with huge version strings.  The varchar 50 assumption in pypicloud doesn't have enough size to cover it.  For example:
https://pypi.python.org/pypi/scipy/0.19.0

Leads to pypicloud issues with the fallback cache:

```
[pid: 12325|app: 0|req: 269/2968] 172.17.0.3 () {38 vars in 857 bytes} [Wed May 24 18:53:49 2017] GET /simple/scipy/ => generated 45864 bytes in 118 msecs (HTTP/1.1 200) 2 headers in 82 bytes (1 switches on core 0)
INFO 2017-05-24 18:53:49,621 [pypicloud.views.api] Caching scipy-0.19.0-cp27-cp27m-macosx_10_6_intel.macosx_10_9_intel.macosx_10_9_x86_64.macosx_10_10_intel.macosx_10_10_x86_64.whl from https://pypi.python.org/simple
Traceback (most recent call last):
  File "/env/local/lib/python2.7/site-packages/pyramid/router.py", line 233, in __call__
    response = self.invoke_subrequest(request, use_tweens=True)
  File "/env/local/lib/python2.7/site-packages/pyramid/router.py", line 208, in invoke_subrequest
    response = handle_request(request)
  File "/env/local/lib/python2.7/site-packages/pyramid_tm/__init__.py", line 120, in tm_tween
    reraise(*exc_info)
  File "/env/local/lib/python2.7/site-packages/pyramid_tm/__init__.py", line 109, in tm_tween
    manager.commit()
  File "/env/local/lib/python2.7/site-packages/transaction/_manager.py", line 131, in commit
    return self.get().commit()
  File "/env/local/lib/python2.7/site-packages/transaction/_transaction.py", line 310, in commit
    reraise(t, v, tb)
  File "/env/local/lib/python2.7/site-packages/transaction/_transaction.py", line 301, in commit
    self._commitResources()
  File "/env/local/lib/python2.7/site-packages/transaction/_transaction.py", line 446, in _commitResources
    reraise(t, v, tb)
  File "/env/local/lib/python2.7/site-packages/transaction/_transaction.py", line 418, in _commitResources
    rm.tpc_begin(self)
  File "/env/local/lib/python2.7/site-packages/zope/sqlalchemy/datamanager.py", line 96, in tpc_begin
    self.session.flush()
[...]

PR #128, opened 2017-05-24
bump version field to 1000 characters

triplepoint:
Addresses issue #126.  It wasn't clear what the best size was for this field, but 1000 characters seems unlikely to be too small.


triplepoint:
Note that this required a cache rebuild to take effect.  Also, running that cache rebuild command with 100k+ packages in the repo takes quite a while (enough to timeout the http connection)

stevearc:
Yeah this seems reasonable. It's known that the cache rebuild can take a while; I haven't put any time into optimizing it. I'll merge this into the 0.5 branch so it's included in the next release

PR #131, opened 2017-08-31
Add additional logging

doublerainbo:
We've been seeing occasional flakes in our pypicloud instance, wanted to add some more logging to try to narrow down if it is happen in the ELB side, dynamoDB access, or S3 side. Have been unable to reproduce in a controlled environment so going to add some logs to try to provide more info

### DIFF
--- a/pypicloud/cache/dynamo.py
+++ b/pypicloud/cache/dynamo.py
@@ -95,6 +95,7 @@ class DynamoCache(ICache):
             connection = DynamoDBConnection.connect(region,
                                                     access_key=access_key,
                                                     secret_key=secret_key)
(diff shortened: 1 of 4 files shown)


doublerainbo:
i'll uhh... look into those failing tests next week 😅

stevearc:
This is quite a lot of logs for one specific part of the code. Could you try deploying this patch just to your server? I'm happy to take a look at any bugs you find with it, but I don't think adding all these logs is a good idea for everyone.

doublerainbo:
the results were not very conclusive. nothing looked very out of place pretty much resorted to adding in retries

PR #135, opened 2017-09-25
Add `pypi.reload_interval` parameter to force cache flushing

apetresc:
Hi Steve!

We're using a somewhat unusual deployment pattern for our pypicloud repositories: we have multiple copies of the frontend running, but they are all backed by the same S3 bucket (for a variety of reasons, mainly having to do with having fine-grained firewall rules for clients). This works fine except for one thing: when a new version of an already-existing package is pushed to one frontend, the other frontends never see it because the configured `ICache` assumes any changes to the underlying storage system must pass through it, unless someone logs in and rebuilds the package list.

The solution we've settled on is to add an optional `pypi.reload_interval` parameter (in seconds). When not set, behavior should be 100% identical to 0.5.5. When it is set, the cache is flushed and `reload_from_storage()` is called on any `fetch()`, `all()`, `distinct()`, `summary()`, and `search()` operations that have been called for the first time after `reload_interval` seconds have elapsed.

This is a nice tradeoff, where a single build will generally only have to pay the reload cost once, but uploading a new package version will eventually propagate on its own.

I've tested this version internally and it appears to work well for our purposes.

### DIFF
--- a/pypicloud/cache/__init__.py
+++ b/pypicloud/cache/__init__.py
@@ -21,7 +21,7 @@ def includeme(config):
     cache_impl = resolver.resolve(dotted_cache)
     kwargs = cache_impl.configure(settings)
     cache = cache_impl(**kwargs)
-    cache.reload_if_needed()
+    cache.reload_if_empty()
     config.add_request_method(partial(cache_impl, **kwargs), name='db',
                               reify=True)
     return cache_impl
--- a/pypicloud/cache/base.py
+++ b/pypicloud/cache/base.py
@@ -44,7 +44,7 @@ class ICache(object):
         has been exceeded.
         """
         elapsed_time = (datetime.now() - ICache.last_reloaded).seconds
-        if self.reload_interval and elapsed_time > self.reload_interval:
+        if elapsed_time >= self.reload_interval > 0:
             LOG.info("Cache is expired (%d seconds). Rebuilding from storage backend..." % elapsed_time)
             self.reload_from_storage()
             LOG.info("Cache repopulated")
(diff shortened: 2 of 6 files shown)


stevearc:
Hi Adrian, thanks for the PR!

You should check out some of the changes in the [1.0 release candidate](http://pypicloud.readthedocs.io/en/1.0/changes.html). The `graceful_reload` option and the S3 notification hooks might be of interest.

The current implementation will probably work fine for low-traffic situations, but it'll cause a [cache stampede](https://en.wikipedia.org/wiki/Cache_stampede) every time `last_reloaded` expires. You could alleviate that by updating `last_reloaded` at the beginning of `reload_from_storage`, but with multiple server processes it could still be an issue. You get much better guarantees if you can rebuild the cache as a completely independent process.

I think the ideal case would be if you can use [S3 notifications](http://pypicloud.readthedocs.io/en/1.0/topics/s3_sync.html#s3-sync) to trigger a cache write, but I understand that setting up Lambda and everything else might be too much of a pain. Can I ask what WSGI server you're using? If you're using uWSGI then it would be easy to hook in to the [cron system](http://uwsgi-docs.readthedocs.io/en/latest/Cron.html) or to use [timers](http://uwsgi-docs.readthedocs.io/en/latest/Signals.html#timers). If not then the easiest thing may be to add a command-line control like `ppc-rebuild-cache` so you can integrate it with cron or anything else.

What do you think?

apetresc:
Thanks @stevearc - the solution in the `1.0` branch is infinitely better. I wasn't aware that there was a 1.0-rc at all!

I'm going to go with generating the lambda bundle with `ppc-build-lambda-bundle` and then creating all the roles and lambdas and notifications within our Terraform setup, which is how we deploy all our AWS resources. Let me know if you'd be interested in a PR with the Terraform config, which I'm sure is a very popular setup for lots of users, even if its just to be used in the documentation :)

Thanks! I'll let you know if I run into any issues with the S3 notification handler.

Closing this issue since this solution supersedes mine.

stevearc:
Yeah, let me know how it goes! I'd definitely be interested to see the terraform config. I'm sure it would be helpful to someone to be able to see what a full setup looks like.

PR #136, opened 2017-09-30
Allow periods in bucket name on pypicloud-make-config

yoichi:
I've tried to use pypicloud with s3 bucket which contains '.'s in its name.
Since pypicloud-make-config doesn't accept s3 bucket name include period,
I have to give dummy bucket name to pypicloud-make-config, and modify output
file by hand. Then pypicloud works well with the configuration file.

This change modifies validate function to accept s3 bucket name contains '.' unless

* start with period
* end with period
* have multiple periods between labels

The condition is described in the s3 official document:
http://docs.aws.amazon.com/AmazonS3/latest/dev/BucketRestrictions.html

Thanks,

### DIFF
--- a/pypicloud/scripts.py
+++ b/pypicloud/scripts.py
@@ -76,6 +76,20 @@ def promptyn(msg, default=None):
             return default
 
 
+def bucket_validate(name):
+    """ Check for valid bucket name """
+    if name.startswith('.'):
+        print "Bucket names cannot start with '.'"
+        return False
+    if name.endswith('.'):
+        print "Bucket names cannot end with '.'"
+        return False
+    if '..' in name:
+        print "Bucket names cannot contain '..'"
+        return False
+    return True
+
+
 def make_config(argv=None):
     """ Create a server config file """
     if argv is None:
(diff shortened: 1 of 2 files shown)


stevearc:
The original reason why I disallowed periods in bucket names was that I tried it once and it caused problems further down the line. Amazon's SSL cert was valid for `foobar.s3.amazonaws.com` but it wasn't valid for `foo.bar.s3.amazonaws.com`. That caused the pip to complain about the installation links and fail to work. Has Amazon changed the behavior? Do the generated links for a bucket with a period in the name have a valid cert?

yoichi:
Thanks for clarify the reason.

I had tested uploading package via API and pip search, but I not with pip install.
Now I've confirmed that pip install does not work with bucket name containing period.
(I've also confirmed there is no problem with bucket name not containing period)

The package URL in such case is something like
https://s3.amazonaws.com/bucket.name.contains.dot/...
and it cause following error:
```
<Error>
  <Code>PermanentRedirect</Code>
  <Message>The bucket you are attempting to access must be addressed using the specified endpoint. Please send all future requests to this endpoint.</Message>
...
```

I found that when I replace domain part "s3.amazonaws.com" to "s3-region_name.amazonaws.com" (where region_name is one of the bucket), then I can download the file.
I don't know well pypicloud implementation. Is this a problem with pypicloud or boto3?
(I'm using pypicloud 1.0.0rc2 since I'm behind http proxy)

yoichi:
Putting "storage.region_name" in config solved my problem (expected package urls are generated).
I've confirmed that we can use pypicloud with s3 bucket contains periods.

stevearc:
Nice, it looks like they changed the generated URL format to use the bucket name in the path instead of the subdomain, which solves this problem. I've merged this commit into the [1.0](https://github.com/stevearc/pypicloud/tree/1.0) branch. It'll go out when 1.0.0 is ready for release.

Thanks for the PR!

stevearc:
This was merged and released in [1.0.0](https://pypi.python.org/pypi/pypicloud/1.0.0)

PR #138, opened 2017-10-14
make pypicloud-make-config python3 compatible

yoichi:
pypicloud-make-config in 1.0 branch didn't work with python3.
This change make it python3 compatible.

* raw_input is renamed, use six.moves.input to support py2 and py3
* decode encrypt_key and validate_key to avoid b'...' output
* decode config.ini.jinja2 to avoid TypeError at Template()

I've tested that python2.7 and python3.6 produce equivalent config.

Regards,


stevearc:
Nice, thanks!

PR #139, opened 2017-10-23
S3: Set ACL for uploaded objects

gmykhailiuta:
We run 2 pypiclouds in different AWS accounts, which share one bucket.
In order to make objects accessible from both accounts, custom canned ACL has to be set when uploading objects ("public-read-write" in our case). Otherwise, AWS default ACL is "private", so when user in account A uploads object to account B's bucket, users in account B can access bucket, but not the objects in it.


stevearc:
Thanks! I merged into 0.5 and 1.0. Will do a new release this weekend

gmykhailiuta:
Awesome! Thanks!

PR #140, opened 2017-11-06
S3 storage config names appear to be invalid

cah-jacksongilman:
Tried to create a config using `storage.aws_secret_access_key` and `storage.aws_access_key_id` but the server could not authenticate with S3.  Changed it to "access_key" and "secret_key" after going through error stacktrace and seeing what the source code looks for.

This also might need to be changed for the dynamodb cache backend...

### DIFF
--- a/doc/topics/storage.rst
+++ b/doc/topics/storage.rst
@@ -67,7 +67,7 @@ startup. If blank, the classic US region will be used.
 
 .. _s3_credentials:
 
-``storage.aws_access_key_id``, ``storage.aws_secret_access_key``
+``storage.access_key``, ``storage.secret_key``
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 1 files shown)


stevearc:
What version of pypicloud are you running? In version 1.0.x we should be using [aws_access_key_id and aws_secret_access_key](https://github.com/stevearc/pypicloud/blob/0768f90d3df1ef34d059d465792cfb0669160999/pypicloud/storage/s3.py#L123-L124). If you're using 0.5.x then we do use [access_key and secret_key](https://github.com/stevearc/pypicloud/blob/4125a426c2113f97ea5910e7c64d5529ebcaa2da/pypicloud/storage/s3.py#L61-L64), but the 0.5 docs also [specify that](http://pypicloud.readthedocs.io/en/0.5.5/topics/storage.html#storage-access-key-storage-secret-key)

cah-jacksongilman:
@stevearc My mistake!  I see now the image we were creating the container with was very stale.  Sorry to inadvertently use a PR to troubleshoot!! :blush:

Fantastic work being done here!

PR #142, opened 2017-11-15
Support for ldap anonymous bind and cert verification

rtkmhart:
Anonymous bind for the service account doesn't work if you specify a username and blank password, so add support for not specifying a service_dn at all.

Also add support for ignoring cert verification on the ldaps protocol which is a valid use case. LDAP is really picky on the hostname and the cert name matching, which breaks when you use DNS aliases.

Also add some debug logging I needed while troubleshooting.

### DIFF
--- a/pypicloud/access/ldap_.py
+++ b/pypicloud/access/ldap_.py
@@ -2,7 +2,7 @@
 import logging
 from collections import namedtuple
 from functools import wraps
-from pyramid.settings import aslist
+from pyramid.settings import aslist, asbool
 
 from .base import IAccessBackend
 from pypicloud.util import TimedCache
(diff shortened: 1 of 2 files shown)


pypicloud/access/ldap_.py
```diff
@@ -141,10 +152,15 @@ def verify_user(self, username, password):
...
-            )
+            if self._service_dn:
+                # bind with the service_dn
+                self._server.simple_bind_s(
+                    self._service_dn,
+                    self._service_password
+                )
+            else:
+                # force a connection without binding
+                self._server.whoami_s()
```
stevearc: Since this "if...else..." logic is used twice, can you factor it out into a method and call it in both places?

pypicloud/access/ldap_.py
```diff
@@ -189,6 +206,8 @@ def group_members(self, group):
         return []  # pragma: no cover
 
     def is_admin(self, username):
+        if not username:
+            return False
```
stevearc: Why was this change needed? It's fine, I'm just curious what was causing is_admin to be called with a falsey username.

Issue #115, opened 2017-03-02
High CPU usage.


luqasz:
Hi.

Thx for your application. It does make a difference when using it. Our CI servers when building, can put a high load on pypicloud. Simultaneous deploys on ~120 instances fails as pypi is overloaded. Any idea what can be done to increase pypicloud performance ?

What we use:
* redis cache
* 25 CI servers
* s3 storage
* c4.xlarge instance
* fallback set to redirect

Maybe my problem somehow refers to #110

luqasz:
Hi.

I've added some caching in nginx and it reduced CPU load significantly.

In http context:
```
uwsgi_cache_path /tmp levels=1:2 keys_zone=pypi_cache:10m max_size=1g inactive=1h;
```
In server context:
```
    location /pypi {
        uwsgi_cache pypi_cache;
        uwsgi_cache_valid 200 5m;
        # cache 302 responses which correspond to redirects to PyPI for packages not on PyPICloud
        uwsgi_cache_valid 302 1h;
        uwsgi_cache_methods GET;
        # distinguish by request_uri
        uwsgi_cache_key $request_uri;
        uwsgi_pass 127.0.0.1:8080;
        include uwsgi_params;
    }
```

Since we do not host a lot of packages, we often return 302.

stevearc:
It's always hard to find the root cause of slowness without a proper profile, but I had to take a shot in the dark I would say maybe the slow cryptography library. If you haven't already done so, try setting [storage.redirect_urls = true](http://pypicloud.readthedocs.io/en/latest/topics/redirect_urls.html#redirect-detail).

Another idea, if you have set `pypi.always_show_upstream = true` then I could see a path where that could cause a lot of slowdown. If it looks like that might be the problem then I could investigate a way to speed that up with caching.

If you can get any sort of profile I'd be very curious to see what appears to be the bottleneck.

luqasz:
How to gather profile ?

stevearc:
You would use [WSGI profiling middleware](http://werkzeug.pocoo.org/docs/0.12/contrib/profiler/)

If you want I threw together a quick patch here: https://github.com/stevearc/pypicloud/commit/210cf411efad2ad33d57ea3be704e7d09bbedc94
You can grab that and add the following to your config file:
```
profile.enable = true
profile.dir = prof
```
Then it should log pstat files to the `prof` directory. You can analyze them with `python -m pstats` or send them to me and I'll take a look

luqasz:
[...]

PR #164, opened 2018-04-13
Added AWS Secrets Manager auth class

koleror:
- Added IMutableJsonAccessBackend to handle auth on any backend that support read/write on a json
- Added AWSSecretsManagerAccessBackend implementing IMutableJsonAccessBackend

### DIFF
--- a/pypicloud/access/aws_secrets_manager.py
+++ b/pypicloud/access/aws_secrets_manager.py
@@ -1,0 +1,181 @@
+""" Backend that defers to another server for access control """
+import boto3
+from botocore.exceptions import ClientError
+import json
+
+from .base import IAccessBackend
+
+"""
+Secret should look like this on AWS:
+
+{
+  "users": {
+    "user1": "password1",
+    "user2": "password2",
+    "user3": "password3",
+    "user4": "password4",
+    "user5": "password5",
+  },
+  "groups": {
+    "admins": [
+      "user1",
+      "user2"
+    ],
+    "group1": [
+      "user3"
+    ]
+  },
+  "packages": {
+      "mypackage": {
+          "groups": {
+              "group1": ["read', "write"],
+              "group2": ["read"],
+              "group3": [],
+          },
+          "users": {
+              "user1": ["read", "write"],
+              "user2": ["read"],
+              "user3": [],
+              "user5": ["read"],
+          }
+      }
+  }
+}
+"""
+
+
+class AWSSecretsManagerAccessBackend(IAccessBackend):
+
+    """
+    This backend allows you to store all user and package permissions in
+    AWS Secret Manager
+
+    """
+
+    def __init__(self, request=None, settings=None, region=None,
+                 secret_id=None, credentials=None, **kwargs):
+        super(AWSSecretsManagerAccessBackend, self).__init__(request, **kwargs)
+        self._settings = settings
+        self.region = region
+        self.secret_id = secret_id
+        self.credentials = {}
+        self.credentials.update(credentials or {})
+
+    @classmethod
+    def configure(cls, settings):
+        kwargs = super(AWSSecretsManagerAccessBackend, cls).configure(settings)
+        kwargs['settings'] = settings
+        kwargs['region'] = settings['auth.region']
+        kwargs['secret_id'] = settings['auth.secret_id']
+        credentials = {}
+
+        if settings.get('auth.access_key'):
+            credentials['aws_access_key_id'] = settings['auth.access_key']
+        if settings.get('auth.secret_key'):
+            credentials['aws_secret_access_key'] = settings['auth.secret_key']
+        if settings.get('auth.session_token'):
+            credentials['aws_session_token'] = settings['auth.session_token']
+        if settings.get('auth.profile_name'):
+            credentials['profile_name'] = settings['auth.profile_name']
+        kwargs['credentials'] = credentials
+        return kwargs
+
+    def _fetch_credentials(self):
+        """ Hit a server endpoint and return the json response """
+        session = boto3.session.Session(**self.credentials)
+        client = session.client(
+            service_name='secretsmanager',
+            region_name=self.region,
+            endpoint_url="https://secretsmanager.{}.amazonaws.com".format(
+                self.region
+            )
+        )
+
+        try:
+            return json.loads(client.get_secret_value(
+                SecretId=self.secret_id
+            )['SecretString'])
+        except ClientError as e:
+            if e.response['Error']['Code'] == 'ResourceNotFoundException':
+                raise Exception(
+                    "The requested secret " + self.secret_id +
+                    " was not found")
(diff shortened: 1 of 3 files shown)


stevearc:
Thanks for the PR! This is exactly the right place for this. Give me just a day or so to go over it

koleror:
Hmm that's an interesting feedback.
I did not know about this service. What's the difference with it?
From what I've seen, it's more or less the same, it stores a encrypted string that can be decrypted.
What we could do on either of the 2 services is add a secret key in the pypi conf, and hash the passwords given by user to compare with passwords that would be hashed the same way on AWS. 
If I do that, I need to provide a route (ideally on Pypi, but not sure it exists) or on AWS to update the password.
What do you think?

tomelliff:
It's basically a precursor to Secrets Manager as people (myself included) were using the SecureString parameter type to hold encrypted secrets for applications to be able to fetch at startup with ECS/Lambda functions.

The main point is that user passwords shouldn't be encrypted and should instead be hashed. And because that's the only thing that's secret here you can use a straight String type instead and not encrypt anything.

You don't need a secret key to hash something, just specify a hashing algorithm, ideally a salt, and some form of work factor or rounds. PyPICloud already hashes passwords for the config and sql auth backends so you can see how that uses passlib to hash things already (although I don't think it provides a salt right now and at a quick glance also seems to be hard-coding the amount of rounds to be used when I was under the impression it calculated an appropriate amount of rounds automatically).

pypicloud/access/aws_secrets_manager.py
```diff
@@ -0,0 +1,181 @@
...
+    def verify_user(self, username, password):
+        credentials = self._fetch_credentials()
+        try:
+            return credentials.get('users', {})[username] == password
+        except KeyError:
+            return False
+
+    def _get_password_hash(self, username):
+        # We don't have to do anything here because we overrode 'verify_user'
+        pass
```
stevearc: As @tomelliff mentioned, it makes more sense to store the passwords hashed. It should be the same as the instructions for using the [config file for auth](http://pypicloud.readthedocs.io/en/latest/topics/access_control.html#user-username). You run `ppc-gen-password` for each of the users instead of putting the raw password into the json blob.

Some background on passlib: the default format that we're using here *does* include a salt, so that's working as intended. The [default number of rounds](https://github.com/stevearc/pypicloud/blob/e9b1d3d284a65f9ad8767b64044c2b1bbeecc315/pypicloud/access/base.py#L17) that we've chosen for pypicloud is based on [the default rounds provided by passlib](https://bitbucket.org/ecollins/passlib/src/4801587cebf01f5037ddc9cd52fc94708559bbfb/passlib/apps.py?at=stable&fileviewer=file-view-default#apps.py-71:85), but was made customizable for #115.

Passlib can choose *secure* defaults for the rounds, but it wouldn't be able to choose *smart* defaults because the rounds are determined at the time of creating the password hash. So in order to do anything smart like choosing a number of rounds that corresponds to a specific number of milliseconds, you would have to be generating the hashes on the same machine that is going to be hosting your webserver. For this and the config file backend, I do not expect that will often be the case.

pypicloud/access/aws_secrets_manager.py
```diff
@@ -0,0 +1,181 @@
...
+                groups.append(group_name)
+        return groups
+
+    def group_members(self, group):
+        credentials = self._fetch_credentials()
+        return list(credentials.get('groups', {}).get(group, []))
+
+    def is_admin(self, username):
+        credentials = self._fetch_credentials()
+        admins = credentials.get('groups', {}).get('admins', [])
```
stevearc: I would recommend putting `admins` at the top level instead of nesting it under `groups`, since it serves a slightly different purpose.

pypicloud/access/aws_secrets_manager.py
```diff
@@ -0,0 +1,181 @@
...
+        users = []
+        for user in credentials.get('users', {}):
+            if username and user != username:
+                continue
+            users.append({
+                'username': user,
+                'admin': user in admins
+            })
+            if username and user == username:
+                break
```
stevearc: You'll need to return a dict here that includes the groups that the user is in. The API could be better, I know. You can find the documentation here https://github.com/stevearc/pypicloud/blob/e9b1d3d284a65f9ad8767b64044c2b1bbeecc315/pypicloud/access/base.py#L412-L418

pypicloud/access/aws_secrets_manager.py
```diff
@@ -0,0 +1,181 @@
...
+    """
+    This backend allows you to store all user and package permissions in
+    AWS Secret Manager
+
+    """
+
+    def __init__(self, request=None, settings=None, region=None,
+                 secret_id=None, credentials=None, **kwargs):
+        super(AWSSecretsManagerAccessBackend, self).__init__(request, **kwargs)
+        self._settings = settings
```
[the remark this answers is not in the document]
koleror: No it's not, thx.

pypicloud/access/aws_secrets_manager.py
```diff
@@ -0,0 +1,181 @@
...
+            )['SecretString'])
+        except ClientError as e:
+            if e.response['Error']['Code'] == 'ResourceNotFoundException':
+                raise Exception(
+                    "The requested secret " + self.secret_id +
+                    " was not found")
+            elif e.response['Error']['Code'] == 'InvalidRequestException':
+                raise Exception("The request was invalid due to:", e)
+            elif e.response['Error']['Code'] == 'InvalidParameterException':
+                raise Exception("The request had invalid params:", e)
```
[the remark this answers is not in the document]
koleror: 👍

pypicloud/access/aws_secrets_manager.py
```diff
@@ -0,0 +1,181 @@
...
+        if settings.get('auth.secret_key'):
+            credentials['aws_secret_access_key'] = settings['auth.secret_key']
+        if settings.get('auth.session_token'):
+            credentials['aws_session_token'] = settings['auth.session_token']
+        if settings.get('auth.profile_name'):
+            credentials['profile_name'] = settings['auth.profile_name']
+        kwargs['credentials'] = credentials
+        return kwargs
+
+    def _fetch_credentials(self):
```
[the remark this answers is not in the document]
koleror: 👍 Hmm I was actually thinking of a cache mechanism, thx :)

pypicloud/access/aws_secrets_manager.py
koleror: I agree, and actually found methods to create and update passwords for a user in the base class (that I didn't find before as I started from the external api class), but I couldn't found the routes or the interface to use them. How do you do that?

Also, if we're not using the password rotation feature from AWS secret manager, I'm thinking of moving the secrets to SSM Parameter Store as it looks cheaper.

pypicloud/access/aws_secrets_manager.py
koleror: agreed

Issue #186, opened 2018-07-16
wrong json redirect


thehesiod:
when navigating to: https://{HOST_NAME}/pypi/{MODULE_NAME}/json on a pypicloud server with redirects enabled, it 3xx's to: https://pypi.org/simple/{MODULE_NAME}/ which is wrong because the ends up being the XML page and not the JSON page.  The redirect needs to be: https://pypi.org/pypi/fiona/json

stevearc:
Fixed in #188

PR #187, opened 2018-07-17
fix json redirects

thehesiod:
fixes https://github.com/stevearc/pypicloud/issues/186

### DIFF
--- a/pypicloud/__init__.py
+++ b/pypicloud/__init__.py
@@ -1,13 +1,12 @@
 """ S3-backed pypi server """
 import calendar
 import datetime
-import io
 import logging
 from pyramid.config import Configurator
 from pyramid.renderers import JSON, render
 from pyramid.settings import asbool
 from pyramid_beaker import session_factory_from_settings
-from six.moves.urllib.parse import urlencode  # pylint: disable=F0401,E0611
+from six.moves.urllib.parse import urlencode, urlparse  # pylint: disable=F0401,E0611
 
 from .route import Root
 from .util import BetterScrapingLocator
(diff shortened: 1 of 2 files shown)


thehesiod:
ugh, what a pain, /simple and /pypi are NOT interchangeable.

Examples
1) https://pypi.org/simple/fiona/json yields 404, whereas https://pypi.org/pypi/fiona/json returns the correct results.
2) https://pypi.org/simple/Fiona/ returns the simple results, whereas https://pypi.org/pypi/fiona/ redirects to https://pypi.org/project/Fiona/ and is the product page.

stevearc:
Thanks for the PR!

Merging this patch as-is will change the behavior of `fallback_url`. If someone had a `fallback_url` with an empty path that was functioning as `/simple`, this will cause their links to be broken now. A couple ideas:

1. If the redirect is happening from a `/<package>/json` endpoint (i.e. `context.name == 'json'`), you could parse the fallback_url and if the last path component is `simple`, switch it to `pypi` then complete the redirect. This is a bit of a hack, but it fixes most people's configurations in-place and doesn't break anything that's currently working, as it'll only affect the `/json` redirects.
or
2. Deprecate `fallback_url` and replace it with a new option, like `redirect_url_base`. Probably also add `redirect_url_simple` and `redirect_url_pypi` in case someone needs to specify a redirect url without `/simple` or `/pypi` in it.

I'm open to other ideas as well, those were just the first two that came to mind.

thehesiod:
ok, thinking about this more:

so pypicloud is support to be re-directing to another pypi server.  The official pypi server is http://pypi.python.org/ (hereby PPYPI)  so all pypi servers should obey to PPYPI's implementation.  The current default pypicloud implementation does not so it needs to be fixed.  pypicloud potentially supports redirecting to non-PYPI prefixed pypi servers so it needs to support that as well.

Given this, and the fact that making it support broken backwards compatibility is pretty ugly, I think the current implementation is almost ok if the major or minor rev is bumped (to signify there's a potential breaking change) and the change in behavior is described as:

1) to forward to a server that behaves like PPYPI server do not specify a path on `fallback_url `. 
2) any path on the `fallback_url` will be treated as a prefix when redirected.

if this is accepted, then I'd change the default to not have a prefix.  If this isn't accepted, and backwards compatibility is paramount then ya I'd introduce yet another configuration var :)

but realistically I think the cleanest is slightly changing the behavior...that's why there's semantic versioning :)

thoughts?

stevearc:
I'm 90% on board with your proposal (it will require a minor version bump), but I do want to keep compatibility with old configs. As long as `fallback_url` continues to behave as expected, I'm fine with changing the expectations and moving to a new config var.

thehesiod:
k, I'll create a new PR soon with the changes as I have some prod builds which points to this branch.

thehesiod:
superseded by https://github.com/stevearc/pypicloud/pull/188 per convo

PR #190, opened 2018-08-16
LDAP admin fix

zarlant:
Iterating over the list returned from AD instead of the admin value specified (which is a string).

Setting the auth.ldap.admin_field to memberOf
Setting the auth.ldap.admin_value to "a given group value"

This fix iterates over the AD values returned instead of iterating over the string admin_value.

### DIFF
--- a/pypicloud/access/ldap_.py
+++ b/pypicloud/access/ldap_.py
@@ -138,8 +138,8 @@ class LDAP(object):
         is_admin = False
         if self._admin_field is not None:
             if self._admin_field in attributes:
-                is_admin = any((val in attributes[self._admin_field]
-                                for val in self._admin_value))
+                is_admin = any((val in self._admin_value 
+                                for val in attributes[self._admin_field]))
 
         return User(username, dn, is_admin)
 


stevearc:
Hey @zarlant, thanks for the PR!

I'm confused about a couple of things. First, you mention that the admin value is a string, but we're very clearly fetching it as a list from the settings.
https://github.com/stevearc/pypicloud/blob/7ed37fc055eeb7df30eb90731dac9ebaf5f5529f/pypicloud/access/ldap_.py#L189
If you're seeing it as a string then there's definitely something going horribly wrong and is likely the root cause of your issue.

Secondly, even if `self._admin_value` was a string, your change is still using the `in` operator. Technically that would work since `'a' in 'a' == True`, but it doesn't make much sense to use it that way.

zarlant:
Ah, I didn't actually see the aslist method and as such did not repro the bug in the same way. The actual issue appears to be that the data is coming back as bytes instead of strings and as such the comparison fails. 

I'll add a new PR to fix that instead.

PR #191, opened 2018-08-27
Update storage.rst

mchalek:
'mornin @stevearc , I was just looking at how the docs for GCS are rendered, and I realized I messed it up a little bit.  This change seems to correctly surface GCS as a separate section from CloudFront in the docs.  Sorry bout that!


stevearc:
Merged into 1.0 branch. Thanks!

PR #192, opened 2018-08-27
Fixing the logic for LDAP reconnects

zarlant:
Without the self inside the try, reconnecting fails to pass the correct data to the wrapped function.


stevearc:
Looks good, will merge into 1.0

PR #193, opened 2018-08-27
Decode bytes so that they match the strings configured in settings

zarlant:


stevearc:
Looks good, will merge into 1.0. I took a while to review this because I figured it was finally time to learn how to set up an LDAP server so I can actually test this thing (the original implementation was a PR). I wanted to hammer some edge cases and make sure that the encoding works as expected; I've been burned before by python unicode issues

PR #206, opened 2019-04-01
Fix ldap admin-value parsing to allow spaces in binding strings

WeeGoSh:
The `aslist` function in pyramid-settings has default `flatten=True`
(https://docs.pylonsproject.org/projects/pyramid/en/latest/api/settings.html) which will make ldap
binding strings with spaces unusable.
Switching the `flatten` flag to `False` will disable cutting the binding strings on spaces and allow the application to
match `admin_value` list against the `admin_field` values correctly.


WeeGoSh:
It might be incorrect to call these binding-strings. These are usually groups via the `memberOf` attribute.

stevearc:
Thanks! Merged into [the 1.0 branch](https://github.com/stevearc/pypicloud/commits/1.0). Will go out in next release

PR #207, opened 2019-04-08
Replace defaukt http://pypi.python.org with https://pypi.org for fall…

fizyk:
…back_base_url

    python package index migrated over a year ago to new domain and new tech stack

    plus the ssl encrypted connection is becoming required


stevearc:
Thanks! I've merged it into the 1.0 branch. Will go out in the next release

Issue #214, opened 2019-05-20
Ability to deny cache_update per repository


mamoit:
I'm running a pypi server as cache for pypi.org (trying not to DoS them :smile:), while also hosting some private repos.
This private repos have some clashes in terms of names with ones that exist on pypi.org.
Since every user has the ability to `update_cache`, to get the latest and greatest version of a package on pypi, they can also pull a private package from pypicloud and accidentally pull a greater version of the homonymous package from pypi.org.
There doesn't seem to be a way to deny/allow to give per package `update_cache` permissions to different users, or is there?
Ideally I'd want to deny `update_cache` entirely on a per package basis.

I've been trying pypicloud for the past weeks and this seems to be an awesome tool, extremely well suited for cloud provider environments, kudos.

stevearc:
It's not possible the way it's currently built, but you could hack that logic in pretty easily, depending on how dirty you're willing to make the hacks 😛

You care about [access.can_update_cache](https://github.com/stevearc/pypicloud/search?q=can_update_cache&unscoped_q=can_update_cache), particularly the two in [simple.py](https://github.com/stevearc/pypicloud/blob/feae61f02d3abd734d771f4a74159b460e98063a/pypicloud/views/simple.py#L231). You have the package available at the time of that call, so you should be able to modify it to do something specific on a per-package basis.

mamoit:
We're thinking about moving the `can_update_cache` inside the `get_fallback_packages` method in order to evaluate it on a per package basis, and thus moving the permission evaluation to a single place.
We'll make a PR if we manage to implement it in a not too hacky way that can actually be re-used by somebody else in a sane way.

Thanks for the tips :+1:

mamoit:
Would it be interesting to get rid of the `can_update_cache` method entirely and create a new `update_cache` permission and using it as you would with `read` and `write`?
Would that require too much of a revamp?
It doesn't seem like it, but just want your 2 cents before diving into that approach.

stevearc:
Hmmmm...yeah I think that would work. That would allow you to set it for specific packages and then if you add a `default_can_update_cache` to match the `default_read` and `default_write` config options, I think that covers all the bases.

mamoit:
We want a way to deny certain packages instead of allowing specific ones (way easier to maintain), so @neuronist came up with the solution above :point_up:
Let us know what you think about the solution.

stevearc:
Linked PR was merged; closing this issue

PR #216, opened 2019-06-05
Add option to disallow fallback for certain packages

neuronist:
This is my hacky attempt at providing the functionality requested on #214.

Let me know if you have any suggestions or if you don't think this is the right approach to tackle this problem.

This is useful if your internal packages shadow names from Pypi and you want to avoid your internal packages getting upgraded to a higher version of a different package present in Pypi.
To use this you just have to add a list of package names in config.ini. E.g. 
```
pypi.disallow_fallback =
  flask
  numpy
```

### DIFF
--- a/pypicloud/access/base.py
+++ b/pypicloud/access/base.py
@@ -68,6 +68,7 @@ class IAccessBackend(object):
         request=None,
         default_read=None,
         default_write=None,
+        disallow_fallback=None,
         cache_update=None,
         pwd_context=None,
         token_expiration=ONE_WEEK,
@@ -76,6 +77,7 @@ class IAccessBackend(object):
         self.request = request
         self.default_read = default_read
         self.default_write = default_write
+        self.disallow_fallback = disallow_fallback
         self.cache_update = cache_update
         self.pwd_context = pwd_context
         self.token_expiration = token_expiration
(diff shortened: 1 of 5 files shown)


mamoit:
You have broken some tests

pypicloud/access/base.py
```diff
@@ -129,6 +132,11 @@ def allowed_permissions(self, package):
                     all_perms[principal] += ("write",)
                 else:
                     all_perms[principal] = ("write",)
+
+        # add fallback permissions
+        for principal in all_perms:
+            if package not in self.disallow_fallback:
+                all_perms[principal] += ("fallback",)
```
stevearc: Slight preference to put the `if` before the `for` to avoid repetitive lookups in the array.

neuronist:
Thanks for the reply and feedback! I will try to formalize this PR this week. A bit loaded with work currently.

pypicloud/access/base.py
neuronist: fixed with 0bc76bb

neuronist:
@stevearc hey man I changed the for loop, fixed the tests, added 2 new ones and also added relevant configuration documentation. :)

I'm a bit confused with the tests and where to put them or if the ones I made are ok. Please give me some feedback to help me navigate your test setup if necessary. :)

Sorry for the number of different commits. :s  
I tried setting up the test environment locally but wasn't able too and was losing too much time on it.

stevearc:
Thanks for the PR! Merged into the [1.0 branch](https://github.com/stevearc/pypicloud/tree/1.0)

neuronist:
Thanks! I'll close this PR then.

PR #225, opened 2019-12-10
Adding "signature_version" to boto3 resource creation. Adds also "Loc…

renehennig:
…ationConstraint" to bucket creation if region_name is set. Otherwise regions like "eu-central-1" won't work properly.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 sudo: required
 language: python
-python: "3.6"
+python: "3.7"
+dist: xenial
 services:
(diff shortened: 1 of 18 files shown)


renehennig:
Hm. I'm pretty sure the broken test are not my fault :D
But still i need to adjust the tests for my changes ...

renehennig:
Thanks for your fix @stevearc. All tests are green again 🎉

stevearc:
Thanks for the PR! Merged and released

PR #226, opened 2019-12-28
Add option to use IAM signer on GCS

### LINKED ISSUE #205: Support for GCS default credentials.
It would be nice if PyPiCloud could be configured to use the default credentials of the environment if none are provided. This is the default of the Google API client if `GOOGLE_APPLICATION_CREDENTIALS` is not set.
I propose altering the GCS authentication path in 2 ways, firstly instead of directly creating the client using `storage.Client.from_service_account_json` instead use the default credentials path, i.e: `storage_client = storage.Client()` and secondly changing the code to set the environment variable if it's not set, i.e `os.environ["GOOGLE_APPLICATION_CREDENTIALS"] = service_account_json_filename` as to maintain backwards compatibility with specifying the credentials file in config rather than environment.

This should allow PyPiCloud when launched on GCP instances or GKE containers that use service-account-assigner to authenticate using granted service account credentials. Aside from being less configuration it has some security benefits as service accounts issued this way are easy to rotate.

SimonStJG:
Based on https://gist.github.com/jezhumble/91051485db4462add82045ef9ac2a0ec
Fixes #205

### DIFF
--- a/doc/topics/storage.rst
+++ b/doc/topics/storage.rst
@@ -201,19 +201,27 @@ form ``storage.<key> = <value>``).
 Settings supported by the S3 backend that are not currently supported by the
 GCS backend are ``server_side_encryption`` and ``public_url``.
 
-This backend requires a service account JSON key file in order to authenticate
-against the GCS API, even when the server is running in Google Cloud Platform (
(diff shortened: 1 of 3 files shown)


stevearc:
Thanks for the PR! Looks good to me

Issue #239, opened 2020-05-08
ImportError: No module named requests


MeanderingCode:
The `1.0.14` release adds an import for `requests` in https://github.com/stevearc/pypicloud/commit/9ca38db2da63ece243fa164c9c81c4af2ca9fc0a#diff-d6cc2aa696b593d885a38547079537a5R2 but `setup.py` does not add it to the `REQUIREMENTS` array here: https://github.com/stevearc/pypicloud/blob/9ca38db2da63ece243fa164c9c81c4af2ca9fc0a/setup.py#L15-L31
I do not know if this is the only place it would need to be added or if it is the only required package missing from installation requirements.

Our servers are failing to start with `ImportError: No module named requests` when uwsgi attempts to load pypicloud with
```
[app:main]
use = egg:pypicloud
```

In the meantime, we are explicitly installing `requests` into our virtualenvs. If we see further errors we will downgrade to 1.0.13.

lopezm1:
We first saw this error while attempting to access the METADATA


```
   File "/srv/pypicloud/lib/python2.7/site-packages/paste/deploy/loadwsgi.py", line 645, in find_egg_entry_point
new_requirements = dist.requires(req.extras)[::-1]
  File "/srv/pypicloud/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2736, in requires
    pkg_resources.require(self.spec)
  File "/srv/pypicloud/lib/python2.7/site-packages/pkg_resources/__init__.py", line 900, in require
    needed = self.resolve(parse_requirements(requirements))
  File "/srv/pypicloud/lib/python2.7/site-packages/pkg_resources/__init__.py", line 794, in resolve
        new_requirements = dist.requires(req.extras)[::-1]
dm = self._dep_map
  File "/srv/pypicloud/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2736, in requires
  File "/srv/pypicloud/lib/python2.7/site-packages/pkg_resources/__init__.py", line 3023, in _dep_map
    dm = self._dep_map
  File "/srv/pypicloud/lib/python2.7/site-packages/pkg_resources/__init__.py", line 3023, in _dep_map
    self.__dep_map = self._compute_dependencies()
  File "/srv/pypicloud/lib/python2.7/site-packages/pkg_resources/__init__.py", line 3032, in _compute_dependencies
        self.__dep_map = self._compute_dependencies()
for req in self._parsed_pkg_info.get_all('Requires-Dist') or []:
[...]

PR #240, opened 2020-05-08
Add missing dependency "requests" to setup.py

MeanderingCode:
9ca38db adds an import for "requests", but it is missing from `setup.py`'s `install_requires`.

Addresses #239


stevearc:
Thanks for the PR!

PR #197, opened 2018-11-13
Summary sent to s3 to be normalized and stripped of non-ASCII characters

fizyk:
Some unicode characters might pose problem for boto, which in case
    of validators package summary 'Python Data Validation for Humans™'
    threw a valdiation error

Stack trace:
```
Parameter validation failed:
                                        Non ascii characters found in S3 metadata for key "summary", value: "Python Data Validation for Humans™.".  
                                        S3 metadata can only contain ASCII characters. 
                                        Traceback (most recent call last):
                                          File "/opt/pypi/lib/python3.5/site-packages/botocore/handlers.py", line 567, in validate_ascii_metadata
                                            value.encode('ascii')
                                        UnicodeEncodeError: 'ascii' codec can't encode character '\u2122' in position 33: ordinal not in range(128)
                                        
                                        During handling of the above exception, another exception occurred:
                                        
                                        Traceback (most recent call last):
                                          File "/opt/pypi/lib/python3.5/site-packages/pyramid/tweens.py", line 39, in excview_tween
                                            response = handler(request)
                                          File "/opt/pypi/lib/python3.5/site-packages/pyramid/router.py", line 156, in handle_request
                                            view_name
                                          File "/opt/pypi/lib/python3.5/site-packages/pyramid/view.py", line 642, in _call_view
                                            response = view_callable(context, request)
                                          File "/opt/pypi/lib/python3.5/site-packages/pyramid/config/views.py", line 181, in __call__
                                            return view(context, request)
                                          File "/opt/pypi/lib/python3.5/site-packages/pyramid/viewderivers.py", line 390, in attr_view
                                            return view(context, request)
                                          File "/opt/pypi/lib/python3.5/site-packages/pyramid/viewderivers.py", line 368, in predicate_wrapper
                                            return view(context, request)
                                          File "/opt/pypi/lib/python3.5/site-packages/pyramid/viewderivers.py", line 439, in rendered_view
                                            result = view(context, request)
                                          File "/opt/pypi/lib/python3.5/site-packages/pyramid_duh/params.py", line 306, in param_twiddler
                                            return fxn(**scope)
                                          File "/opt/pypi/lib/python3.5/site-packages/pypicloud/views/simple.py", line 37, in upload
                                            version=version, summary=summary)
                                          File "/opt/pypi/lib/python3.5/site-packages/pypicloud/cache/base.py", line 118, in upload
                                            self.storage.upload(new_pkg, data)
[...]

PR #241, opened 2020-05-11
Adding Azure Blob Storage support

terrycain:
Added tests, so it seems the way most of the azure sdk's test is by using vcrpy, which doesn't seem to work on < py36.

### DIFF
--- a/doc/topics/storage.rst
+++ b/doc/topics/storage.rst
@@ -320,19 +320,19 @@ Set ``pypi.storage = azure-blob`` OR ``pypi.s3 = pypicloud.storage.AzureBlobStor
 
 A few key, required options are mentioned below.
 
-``storage.azure_storage_account_name``
+``storage.storage_account_name``
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 **Argument:** string
 
 The name of the Azure Storage Account.
 
-``storage.azure_storage_account_key``
+``storage.storage_account_key``
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 **Argument:** string
 
 A valid access key, either key1 or key2.
 
-``storage.azure_storage_container_name``
+``storage.storage_container_name``
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 **Argument:** string
 
--- a/doc/topics/storage.rst
+++ b/doc/topics/storage.rst
@@ -311,3 +311,29 @@ the default storage class of the bucket, if the bucket is preexisting, or
 Sign blobs using IAM backed signing, rather than using GCP application credentials.
 The service account used needs to have ``iam.serviceAccounts.signBlob`` on the storage
 bucket.  This is available as part of ``roles/iam.serviceAccountTokenCreator``.
+
+Azure Blob Storage
+------------------
+This option will store your packages in a container in Azure Blob Storage.
+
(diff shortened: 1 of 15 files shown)


terrycain:
Coverage is still unhappy about a -0.04% decrease, but I think the tests are fine

doc/topics/storage.rst
```diff
@@ -311,3 +311,29 @@ the default storage class of the bucket, if the bucket is preexisting, or
...
+
+Azure Blob Storage
+------------------
+This option will store your packages in a container in Azure Blob Storage.
+
+Set ``pypi.storage = azure-blob`` OR ``pypi.s3 = pypicloud.storage.AzureBlobStorage``
+
+A few key, required options are mentioned below.
+
+``storage.azure_storage_account_name``
```
stevearc, line 323: For consistency with the other storage options, you don't have to include the storage type in the option name. The storage type is specified in the `pypi.storage = azure-blob` argument, so all of the `storage.` arguments will apply to the azure blob backend

pypicloud/storage/azure_blob.py
```diff
@@ -0,0 +1,180 @@
+""" Store packages in S3 """
```
stevearc, line 1: copypasta :)

pypicloud/storage/azure_blob.py
```diff
@@ -0,0 +1,180 @@
...
+            )
+        return package.data["path"]
+
+    def upload(self, package, datastream):
+        path = self.get_path(package)
+
+        metadata = package.get_metadata()
+        metadata["name"] = package.name
+        metadata["version"] = package.version
+        self._normalize_metadata(metadata)
```
stevearc, line 144: Do you know if this step is actually required for azure-blob storage? This was added in #197 and seems to be because of a limitation of the AWS REST API (it doesn't support non-ascii text). If this *is* also required by azure, could you move the method into the util.py file so both storage classes can use the same function?

pypicloud/storage/azure_blob.py
```diff
@@ -0,0 +1,180 @@
...
+                )
+
+    def delete(self, package):
+        path = self.get_path(package)
+        blob_client = self.container_client.get_blob_client(blob=path)
+        blob_client.delete_blob()
+
+    def check_health(self):
+        try:
+            list(self.container_client.list_blobs(name_starts_with=""))
```
stevearc, line 167: This looks like an expensive operation. Is there anything simpler we could do here? All this is used for is to check that the connection works and the auth credentials are valid.

pypicloud/storage/azure_blob.py
terrycain: :D

pypicloud/storage/azure_blob.py
terrycain: Yep it gets very unhappy when you pass it utf8 characters, will move

pypicloud/storage/azure_blob.py
terrycain: Good shout, will change it to looking for properties of a singular resource, as long as we get a response it'll be good.

terrycain:
Cool fixed all that :)

stevearc:
Awesome, thanks! Merged into the 1.0 branch and it'll go out in the next release

terrycain:
Cool, any rough timescale on that :)

stevearc:
You can expect it by the end of the next weekend

terrycain:
Cool thanks

PR #243, opened 2020-05-19
Drop py27

terrycain:
Drops Python 2.7 and adds some basic type hints where its easy to do so.

Probably best off looking at the diff's between the individual commits.

Might want to look at running `isort` to sort the imports at some point in the future

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -43,3 +43,4 @@ _build
 _local.db
 *.ini
 .idea
+.mypy_cache
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,6 @@ services:
   - postgresql
   - docker
 env:
-  - TOXENV=py27
   - TOXENV=py38
   - TOXENV=lint
 matrix:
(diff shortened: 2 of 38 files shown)


stevearc:
Looking good so far! Is this ready for merge or did you have more changes you wanted to make?

terrycain:
Go ahead and merge, there is some more stuff I want to add but I'll do that in a separate pr when I have the energy :D, mainly just adding in more type hints, completely non-functional aesthetic stuff

stevearc:
Created a 1.1 branch and merged this into that. Also added `mypy` to the tox config so we catch type errors in CI. I also made a branch where I added isorting, but I'm leaving that unmerged for now so it doesn't cause you any merge conflicts. Will merge it in to 1.1 after your PRs are all in.

terrycain:
Nice I'll rebase my hash support pr onto 1.1, will probably finish that one before Monday.

stevearc:
Don't know why this never closed, but it's all done and merged now

PR #244, opened 2020-05-19
Added package hashing logic

### LINKED ISSUE #222: pip package hash support
when using tools like poetry they generate lock files with the hashes of the modules.  I've noticed that when it hits our pypicloud server it doesn't record any hashes (https://github.com/sdispater/poetry/issues/1553), so presumably pypicloud is missing some feature.  I haven't had a chance to look into why this is yet.  Presumably some sort of metadata API/part of response is missing

terrycain:
Adds logic to calculate package hashes on upload, adds it to file metadata as well so if the database is rebuilt, re-hashing is not needed.

New option `pypi.calculate_package_hashes` added to control whether to calculate package hashes. Enabled by default as that means the JSON and simple API will act more like the real PyPI.

Depends on #243 
Fixes #222

## TODO:
* [x] add hashing logic
* [x] ~test for sha256 url fragment in simple endpoint~
* [x] test for hashes in json output
* [x] test for hash calculation on upload
* [x] test for hash retrieval on list
* [x] update docs about `pypi.calculate_package_hashes` option
* [x] Create Azure and S3 migration script (havent used GCE so someone else can create that when needed)

### DIFF
--- a/doc/topics/configuration.rst
+++ b/doc/topics/configuration.rst
@@ -98,6 +98,15 @@ Only used when ``pypi.fallback = cache``. This is
 the list of groups that are allowed to trigger the operation that fetches
 packages from ``fallback_base_url``.  (default ['authenticated'])
 
+``pypi.calculate_package_hashes``
+**Argument:** bool, optional
+
+Package SHA256 and MD5 hashes are now calculated by default when a package is
+uploaded. This option enables or disables the hash calculation (default true)
+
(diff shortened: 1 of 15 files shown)


terrycain:
@stevearc Ready for a review when you get a chance.

stevearc:
Looks good! Will cut a 1.1 release by EOW

terrycain:
Awesome thanks

PR #253, opened 2020-07-23
Add flag to pypicloud-gen-password for reading password from stdin

viktorradnai:
This feature would be used for scripting the password hash creation. The specific use case I have in mind is to generate the hashed password from a Kubernetes secret using an init container before starting pypicloud. 

The equivalent functionality is also available with the `mkpasswd` command which is part of the `whois` package, (`mkpasswd -m sha-512 -s -R 20500`) but it makes sense for the `pypicloud-gen-password` utility to support it as well.


stevearc:
Looks good to me! Merged into the 1.1 branch.

viktorradnai:
Thanks, closing PR.

Issue #254, opened 2020-08-14
s3: underscore should not be used in metadata


everpcpc:
some of `METADATA_FIELDS = ["requires_python", "summary", "hash_sha256", "hash_md5"]` names contain underscore.
In a upload request to s3 storage, the request header would be for example `x-amz-hash_sha256`, which is acturally not a valid http header, and may cause inconsistence in storage.

It may be better to convert them to `hash-sha256` when stored to s3, and change back to `hash_sha256` in `Package.read_metadata`.

PR #255, opened 2020-08-14
fix: store metadata in s3 with '-' instead of '_'

everpcpc:
resolve: https://github.com/stevearc/pypicloud/issues/254

### DIFF
--- a/pypicloud/models.py
+++ b/pypicloud/models.py
@@ -74,6 +74,10 @@ class Package(object):
             value = blob.get(field)
             if value:
                 metadata[field] = value
+                continue
+            value = blob.get(field.replace("_", "-"))
+            if value:
+                metadata[field] = value
         return metadata
 
     def get_metadata(self):
--- a/pypicloud/storage/azure_blob.py
+++ b/pypicloud/storage/azure_blob.py
@@ -132,7 +132,7 @@ class AzureBlobStorage(IStorage):
         metadata = package.get_metadata()
         metadata["name"] = package.name
         metadata["version"] = package.version
-        normalize_metadata(metadata)
+        metadata = normalize_metadata(metadata)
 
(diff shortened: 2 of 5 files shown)


stevearc:
Hi, thanks for the PR! I'm not sure why this change needs to be made, though. The usage of http headers is more of an implementation detail of the API, and I don't think that we should be concerned with it. According to [the S3 documentation](https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingMetadata.html#object-key-guidelines-safe-characters), underscores are considered "safe".

Have you been seeing any errors with the metadata storage? If so, then that's worth bringing it up to the boto library.

everpcpc:
Hi, thanks for the response. Actually I'm not using `aws s3`, but `Tencent cos` which implements s3 api. It could not handle underscore header properly, and will respond with a `SignatureDoesNotMatch`  error.

[the S3 Document](https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingMetadata.html#object-key-guidelines-safe-characters) is about the object key naming, but not metadata key naming. Object key dose not appear in header so it would not cause such problem. And [the metadata document](https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingMetadata.html#UserMetadata) only talks about the value, but did not mention how the key should be formated.

However, I think it's better to follow the http standard to avoid potential issues, since so many other s3 implements could not handle this kind of problem well such as https://github.com/localstack/localstack/issues/459.

As for boto library, I think it should not be responsible for this because metadata is a user defined value, automatically changing user input would cause unexpected behavior. Just like `normalize_name` for the object key, the metadata key should also be normalized by ourselves.

stevearc:
Thanks for the explanation! Given that you're actually experiencing issues and it seems like it could be a common problem, I'm happy to merge. Thanks for the fix!

PR #260, opened 2020-10-06
Fixed content type typo

pepastach:
This pull request fixed a typo in response content type. It seems have been introduced with `stream_files` support.


stevearc:
Thanks for the fix!

PR #262, opened 2020-10-20
Fix #261: Add service account email config to GCS storage

### LINKED ISSUE #261: Give way to provide explicit service account email for GCS IamSigner, or use the existing service account
## What?
When using a specific gcp service account for gcs
it is picking default GCE service account for signing links
instead of using the provided service account

## How?
Problematic lines : 
https://github.com/stevearc/pypicloud/blob/3015160ed6fd7f8ffc3a3e4513e7fb91dd340921/pypicloud/storage/gcs.py#L171-L173

As per
https://github.com/googleapis/google-auth-library-python/blob/6407258956ec42e3b722418cb7f366e5ae9272ec/google/auth/compute_engine/credentials.py#L168-L170
And
https://github.com/googleapis/google-auth-library-python/blob/6407258956ec42e3b722418cb7f366e5ae9272ec/google/auth/compute_engine/credentials.py#L205-L209

It will always pick up the default service account if we dont mention the email in constructor.

## Docs also assume that it uses the same provided gcs service account as json:
https://github.com/stevearc/pypicloud/blame/master/doc/topics/storage.rst#L222-L224

## Why?
It is very difficult to control the default workload service account.
Which is returned by rest call `http://metadata.google.internal/computeMetadata/v1/instance/service-accounts/default/?recursive=true`
In systems like kubernetes, this can be done by several hidden annotations. In systems like docker-compose / any k8s engine apart from GKE, this will fail. 
Ultimately we are creating a "works in my laptop/environment" situation again. 
Explicit configs are always heartthrobs

stevearc:
Fix #261

### DIFF
--- a/pypicloud/storage/gcs.py
+++ b/pypicloud/storage/gcs.py
@@ -29,6 +29,7 @@ class GoogleCloudStorage(ObjectStoreStorage):
         service_account_json_filename=None,
         project_id=None,
         use_iam_signer=False,
+        service_account_email=None,
         **kwargs
     ):
         super(GoogleCloudStorage, self).__init__(request=request, **kwargs)
@@ -37,6 +38,7 @@ class GoogleCloudStorage(ObjectStoreStorage):
         self._bucket_factory = bucket_factory
         self.service_account_json_filename = service_account_json_filename
         self.use_iam_signer = use_iam_signer
+        self.service_account_email = service_account_email
 
         if self.public_url:
             raise NotImplementedError(
@@ -78,6 +80,9 @@ class GoogleCloudStorage(ObjectStoreStorage):
             "service_account_json_filename": service_account_json_filename,
             "project_id": settings.get("storage.gcp_project_id"),
             "use_iam_signer": asbool(settings.get("storage.gcp_use_iam_signer", False)),
+            "service_account_email": settings.get(
+                "storage.service_account_email", False
+            ),
             "bucket_factory": lambda: cls.get_bucket(bucket_name, settings),
         }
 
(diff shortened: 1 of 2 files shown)


pPanda-beta:
Hi @stevearc 
As per https://github.com/googleapis/google-auth-library-python/blob/6407258956ec42e3b722418cb7f366e5ae9272ec/google/auth/compute_engine/credentials.py#L205-L209

this will fix the problem of #261 

On top of that can we have 2-3 lines of code to use the existing service account's email?

pypicloud/storage/gcs.py
```diff
@@ -78,6 +80,9 @@ def _subclass_specific_config(cls, settings, common_config):
             "service_account_json_filename": service_account_json_filename,
             "project_id": settings.get("storage.gcp_project_id"),
             "use_iam_signer": asbool(settings.get("storage.gcp_use_iam_signer", False)),
+            "service_account_email": settings.get(
+                "storage.service_account_email", False
```
pPanda-beta, line 84: can we name it "iam_signer_service_account_email" ?
People may get confused that this is the same service account which is used to access GCS. 

Ideally there are two service accounts
1. Svc account to access gcs
2. Svc account to sign gcs links

(As a fallback I was suggesting to reuse service account # 1 , when # 2 is not explicitly provided.)

Logic should be
1. Use json file for gcs service acount
2. use service account email "iam_signer_service_account_email" for signing.
   if not provided then fallback to the email of json file

stevearc:
@pPanda-beta Sorry for slow iteration; I've been traveling. Does this now handle your use case?

pPanda-beta:
@stevearc 
No worries, this looks good. 
Now the order of config is iam_signer_service_account_email > svc_account.json($.client_email) .
Seem perfect to me. 
Thank you, and can we go ahead and have merge plz?

Issue #266, opened 2020-11-12
can't compare offset-naive and offset-aware datetimes


switchtrue:
I'm using pypicloud with gcs storage and redis cache. When clicking "Rebuild package list" in the admin UI I'm getting `TypeError: can't compare offset-naive and offset-aware datetimes` (stacktrace below).

It's being raised by [this line of code](https://github.com/stevearc/pypicloud/blob/master/pypicloud/cache/redis_cache.py#L269) but I can't quite figure out what that's trying to achieve, nor how it's ended up in this situation.

```
ERROR 2020-11-12 08:12:33,355 [pypicloud.views] can't compare offset-naive and offset-aware datetimes
Traceback (most recent call last):
  File "/opt/pypicloud/venv/lib/python3.8/site-packages/pyramid/tweens.py", line 41, in excview_tween
    response = handler(request)
  File "/opt/pypicloud/venv/lib/python3.8/site-packages/pyramid/router.py", line 147, in handle_request
    response = _call_view(
  File "/opt/pypicloud/venv/lib/python3.8/site-packages/pyramid/view.py", line 683, in _call_view
    response = view_callable(context, request)
  File "/opt/pypicloud/venv/lib/python3.8/site-packages/pyramid/config/views.py", line 188, in attr_view
    return view(context, request)
  File "/opt/pypicloud/venv/lib/python3.8/site-packages/pyramid/config/views.py", line 214, in predicate_wrapper
    return view(context, request)
  File "/opt/pypicloud/venv/lib/python3.8/site-packages/pyramid/viewderivers.py", line 325, in secured_view
    return view(context, request)
  File "/opt/pypicloud/venv/lib/python3.8/site-packages/pyramid/viewderivers.py", line 436, in rendered_view
    result = view(context, request)
  File "/opt/pypicloud/venv/lib/python3.8/site-packages/pyramid/viewderivers.py", line 116, in _class_requestonly_view
    response = getattr(inst, attr)()
  File "/opt/pypicloud/venv/lib/python3.8/site-packages/pypicloud/views/admin.py", line 26, in rebuild_package_list
    self.request.db.reload_from_storage()
  File "/opt/pypicloud/venv/lib/python3.8/site-packages/pypicloud/cache/redis_cache.py", line 269, in reload_from_storage
    if package.last_modified > summary["last_modified"]:
TypeError: can't compare offset-naive and offset-aware datetimes
```

switchtrue:
[...]

Issue #246, opened 2020-06-02
Cache issue with packages that have same name/version but different py version


yanovs:
My coworkers and I have run across an issue when trying to gracefully reload the SQLite-backed cache: when a package has the same name and version, but different python version compatibility (e.g., [`docutils==0.15.2` was released this way](https://pypi.org/project/docutils/0.15.2/#files)), the server only registers one of the packages. If we set `graceful_reload=False`, the issue goes away, but we'd obviously prefer to use the graceful reload.

Here's a branch with both a simple test case that fails in master and a potential fix: https://github.com/yanovs/pypicloud/pull/1. FYI, I'm running this patched version of the server, and things look fine so far.

The potential fix is to have `pypicloud.models.Package` hash and compare based on `self.filename` rather than `self.name` and `self.version` (e.g., then `docutils-0.15.2-py2-none-any.whl` and `docutils-0.15.2-py3-none-any.whl` don't get collapsed in the set).

However, this is obviously a change to a core class of the library, so I was wondering what people thought. (Is `filename` always non-`None`? Would it be better to parse the other parts of the wheel name and include them in the hashing/comparing tuple?)

Thanks!

stevearc:
Oh wow I must have missed this back when you submitted it. My apologies!

Turns out you're completely correct and comparing the filenames is the solution to that problem. This ended up getting reported again later and eventually fixed by #267

PR #267, opened 2020-11-13
Fixing an issue with graceful reload

stevearc:
The reload logic was treating packages as the same if their names and
versions matched. This breaks down if a package has multiple builds for
the same name + version (e.g. wheel, tar.gz, windows, mac, linux, etc).
The solution is to compare the filenames to judge equality instead of
comparing name + version.

### DIFF
--- a/pypicloud/models.py
+++ b/pypicloud/models.py
@@ -90,13 +90,17 @@ class Package(object):
         return metadata
 
     def __hash__(self):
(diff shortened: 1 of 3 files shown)


switchtrue:
I can confirm that this fixes the issues I described on #266 and here. i suspect this is also the same as #246

PR #269, opened 2020-12-15
feat: add more info in simple package version json api

everpcpc:
for generating badge with shields.io from api like pypi.org


stevearc:
Looked over current state and the direction seems fine. LMK when you want a full review

everpcpc:
@stevearc PTAL

stevearc:
Looks good to me!

Issue #271, opened 2021-01-05
Upload to Azure Blob fails: The metadata specified is invalid


alanjcastonguay:
Backed by (what I think is) a fairly straightforward azure blob storage configuration, attempting to upload with `pypicloud==1.1.7` fails with `The metadata specified is invalid. It has characters that are not permitted.`.

The wheel uploaded doesn't seem to matter, but I reproduced following using `requests-2.25.1-py2.py3-none-any.whl` downloaded with `python3 -m pip download requests`.

<details>
  <summary>Logged Exception</summary>
    ```
ERROR 2021-01-05 19:48:08,520 [pypicloud.views] The metadata specified is invalid. It has characters that are not permitted.
RequestId:adc1b9d4-501e-002d-549b-e355a3000000
Time:2021-01-05T19:48:08.5013357Z
ErrorCode:InvalidMetadata
Error:None
Traceback (most recent call last):
  File "/usr/local/lib/python3.5/dist-packages/azure/storage/blob/_upload_helpers.py", line 106, in upload_block_blob
    **kwargs)
  File "/usr/local/lib/python3.5/dist-packages/azure/storage/blob/_generated/operations/_block_blob_operations.py", line 222, in upload
    raise models.StorageErrorException(response, self._deserialize)
azure.storage.blob._generated.models._models_py3.StorageErrorException: Operation returned an invalid status 'The metadata specified is invalid. It has characters that are not permitted.'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
... snip
  File "/usr/local/lib/python3.5/dist-packages/pypicloud/views/simple.py", line 40, in upload
    requires_python=requires_python or None,
  File "/usr/local/lib/python3.5/dist-packages/pypicloud/cache/base.py", line 149, in upload
    self.storage.upload(new_pkg, data)
  File "/usr/local/lib/python3.5/dist-packages/pypicloud/storage/azure_blob.py", line 139, in upload
    blob_client.upload_blob(data=datastream, metadata=metadata)
  File "/usr/local/lib/python3.5/dist-packages/azure/core/tracing/decorator.py", line 83, in wrapper_use_tracer
    return func(*args, **kwargs)
  File "/usr/local/lib/python3.5/dist-packages/azure/storage/blob/_blob_client.py", line 539, in upload_blob
    return upload_block_blob(**options)
  File "/usr/local/lib/python3.5/dist-packages/azure/storage/blob/_upload_helpers.py", line 158, in upload_block_blob
[...]

PR #272, opened 2021-01-08
fix: do not normalize metadata name for azure

everpcpc:
fix: https://github.com/stevearc/pypicloud/issues/271


stevearc:
Seems fine to me

PR #293, opened 2021-12-10
Fix formatting in configuration docs

yoganandc:
option `pypi.calculate_package_hashes` was formatted incorrectly


stevearc:
LGTM, thanks!

PR #303, opened 2022-06-21
Stream files from storage backend chunk by chunk

ddelange:
builds upon #302

ref https://github.com/stevearc/pypicloud/pull/302#discussion_r902518060

after this PR, the lib:
- does not need to read the whole package from db into memory to perform `request.db.storage.open`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -13,6 +13,7 @@ CHANGES = re.sub(r":ref:`(.*?) <.*>`", r"\1", CHANGES)
 
 REQUIREMENTS_TEST = open(os.path.join(HERE, "requirements_test.txt")).readlines()
 REQUIREMENTS = [
+    "smart_open[http]",
     "boto3>=1.7.0",
     # beaker needs this
     "cryptography",
--- a/pypicloud/storage/azure_blob.py
+++ b/pypicloud/storage/azure_blob.py
@@ -2,15 +2,13 @@
 import logging
 import os
 import posixpath
-from contextlib import contextmanager
 from datetime import timedelta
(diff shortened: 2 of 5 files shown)


setup.py
```diff
@@ -13,6 +13,7 @@
 
 REQUIREMENTS_TEST = open(os.path.join(HERE, "requirements_test.txt")).readlines()
 REQUIREMENTS = [
+    "smart_open",
```
stevearc, line 16: Could you change this to `smart_open[http]`? We already depend on requests, but I think it's "more correct" to explicitly specify the dependency. On that note, could you also add the other extras to the `EXTRAS` lists below?

stevearc (changes requested):
Looks good! One small request, and also could you update the `files.py` storage backend? I believe that the `closing` wrapper needs to be removed from `open`.

setup.py
ddelange: would you like to introduce s3 in EXTRAS? my minimally invasive suggestion would be to add s3,http extras here, and add gcs&azure extras to the EXTRAS

stevearc: Yep, that sounds good 👍

ddelange:
rebased and updated files.py & setup.py 👍

ddelange:
wondering why the `closing` didn't break VI 🤔

setup.py
ddelange: done in #304 (and only http extra in this PR to strictly separate the features)

stevearc (approved):
Awesome, thanks!

PR #309, opened 2022-08-10
Add redis cluster support

### LINKED ISSUE #308: Support redis-cluster as a cache option
Hello!

I think it would be great if pypicloud would support redis-cluster, in order to work with high load and high performant setups.

The changes needed are adding an option to the redis configuration whether or not the redis is clustered, and if so - in the code, use the RedisCluster class instead of StrictRedis, as in newer versions of redis python package there is support for Redis in clustered mode.
The interface of the classes is the same, and the code won’t need to change much.

Would love for the support in order to create huge pypicloud instances holding hundred thousand of packages.

Thanks!

YannayH:
Closes #308


stevearc:
Excellent work!

PR #310, opened 2022-08-12
Added tracking of package uploader

YannayH:
Added feature of tracking the user who uploaded a package in the database and in the cache.
The user is displayed as an additional column in the package page in the UI.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,3 +46,6 @@ _local.db
 .idea
 .mypy_cache
 .direnv/
+.tox
+.local
+.pylint.d
--- a/pypicloud/cache/base.py
+++ b/pypicloud/cache/base.py
@@ -98,6 +98,7 @@ class ICache(object):
         name: Optional[str] = None,
         version: Optional[str] = None,
         summary: Optional[str] = None,
+        uploader: Optional[str] = None,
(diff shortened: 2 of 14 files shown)


stevearc:
Thanks for the PR! I don't have time to do a full review right now, but I'll get to it as soon as I'm able.

stevearc:
Apologies, but I'm going off the grid for about a week and don't have time to do a proper review before then. This is still on my todo list, and once I get back I'll get to it as soon as possible

stevearc (changes requested):
In general, I'm on board with this change. I think that the uploading user is a reasonable piece of metadata to add to the packages. Structurally, this PR is adding the uploader directly to the package model classes as a new field, but I'd prefer the approach of putting the uploader into the "metadata" dict on the package for two reasons:

1. Decreases surface area of the change. Requires fewer changes across all the different models, which makes it less likely that there's an edge case missed somewhere.
2. Doesn't change the schema of the backends. As-is, anyone using the SQL cache backend would have to perform a migration to add the new column.

If you have strong reasons for doing it this way instead of using the metadata dict, I could be persuaded.

YannayH:
Closing for PR #312, using your suggested method there.

PR #312, opened 2022-08-26
Added uploader as package metadata

yannayh:
As suggested in PR #310 - trying a different approach using metadata fields instead of adding a new attribute.

### DIFF
--- a/pypicloud/static/partial/package.html
+++ b/pypicloud/static/partial/package.html
@@ -63,7 +63,7 @@
               </button>
             </td>
             <td>
-              {{ package.metadata.uploader !== null ? package.metadata.uploader : "&lt;unknown&gt;" }}
+              {{ package.metadata.uploader != null ? package.metadata.uploader : "&lt;unknown&gt;" }}
             </td>
           </tr>
         </tbody>
--- a/pypicloud/static/partial/package.html
+++ b/pypicloud/static/partial/package.html
@@ -43,6 +43,7 @@
             <th>Package</th>
             <th>Version</th>
             <th>Uploaded</th>
+            <th>Uploaded By</th> 
           </tr>
         </thead>
         <tbody>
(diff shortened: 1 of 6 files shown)


pypicloud/static/partial/package.html
```diff
@@ -61,6 +62,9 @@ <h2>{{ package_name }}</h2>
                 Delete
               </button>
             </td>
+            <td>
+              {{ package.metadata.uploader !== null ? package.metadata.uploader : "&lt;unknown&gt;" }}
```
stevearc, line 66: fix: You'll want to use `!=` here so that it shows `<unknown>` when the package doesn't have an uploader in the metadata

stevearc (changes requested):
One small change requested

pypicloud/static/partial/package.html
YannayH: Done :)
I tested it before using `!=`, and it worked, probably didn't re-check after changing to `!==` as I saw it's more correct to use in JS.
Restored to `!=~

stevearc: Yeah in JS `!==` is generally preferred as it does what `!=` does in most other languages, but most people are willing to make an exception for `!= null` because it's a useful shorthand for "is not null or undefined".

yannayh: Good to know :)
Any more changes you would like me to add?

stevearc:
Thanks! Small request for future PRs: don't modify and force-push commits that have already been reviewed. It makes it harder for the reviewer to see what's changed at each modification.

yannayh:
Hey, can you release a new version including this feature and build the docker image?

stevearc:
Yeah, I was just giving the other PRs a chance to land before releasing. I'll do a release this weekend

yannayh:
Yeah I probably won’t have time to address my other PR in the following week or so unfortunately.

PR #314, opened 2022-08-27
processes metedata continuous spaces

QSummerY:
As suggested with issue https://github.com/boto/botocore/issues/2409
Upload may fail when there is a continuous space

### DIFF
--- a/pypicloud/util.py
+++ b/pypicloud/util.py
@@ -80,7 +80,7 @@ def normalize_metadata_value(value: Union[str, bytes]) -> str:
         value = value.decode("utf-8")
     if isinstance(value, str):
         value = "".join(c for c in unicodedata.normalize("NFKD", value) if ord(c) < 128)
-    return value
+    return re.sub(r"\s+", " ", value)
 
 
 def normalize_metadata(metadata: Dict[str, Union[str, bytes]]) -> Dict[str, str]:


stevearc (changes requested):
The linked issue is talking about leading/trailing whitespace, not multiple space characters in a row. This seems like a `value.strip()` would be more appropriate?

QSummerY:
SignatureDoesNotMatch occurs when I try to upload. After local verification, the signature parsing will be affected if continuous Spaces appear. Non-ascii characters in metadata are removed from the normalize_metadata_value method, after removing the continuous space may come up.
This linked issue is similar to mine, and I also tried the situation with Spaces at the beginning and the end, but the upload is normal, only the continuous space will be affected

pypicloud/util.py
```diff
@@ -80,7 +80,7 @@ def normalize_metadata_value(value: Union[str, bytes]) -> str:
         value = value.decode("utf-8")
     if isinstance(value, str):
         value = "".join(c for c in unicodedata.normalize("NFKD", value) if ord(c) < 128)
-    return value
+    return re.sub(r"\s+", " ", value)
```
stevearc, line 83: This will replace all whitespace characters with `" "`. Could we instead do 
```python
re.sub(r"  +", " ", value)
```
to only replace instances of two-or-more spaces with a single space?

stevearc (changes requested):
Out of curiosity, which object storage provider is giving you this signature error?

pypicloud/util.py
QSummerY: I tried the presence of other whitespace characters, the same problem would occur if \t existed, and the presence of \n and \r would succeed, but the characters after the whitespace were removed. 
And observing that only X-AMZ-meta-summary affects the request, the values processed include the following.
```
'x-amz-meta-hash-sha256': '71eccb33ac8b2584c86f36f8ebfbe72c0a98022b576f411ab05773343e4e2cac', 
 'x-amz-meta-hash-md5': 'f562366338d015df143c7afacabdee44', 
 'x-amz-meta-summary': ' Python3\tweb , WSGI, Web ', 
 'x-amz-meta-name': 'frontgate', 'x-amz-meta-version': '0.6.39'
```
So I think it's going to be implemented the same way as before.

QSummerY: It's the COS of Tencent Cloud.

stevearc: Okay, last request:
This is going to replace all kinds of whitespace with spaces. It would be better if we could preserve the original whitespace if it's not causing issues. I believe this should do it
```python
re.sub(r'(\s)\s+', r'\1', value)
```

stevearc (changes requested):
One last request before merging

pypicloud/util.py
QSummerY: Okay, that looks better

stevearc: Wow COS really doesn't like whitespace, huh? I guess this is fine. The worst that we expect is some mangling of the summary formatting.

stevearc (approved):

PR #316, opened 2022-09-11
Fixed sort in when using redis cache

yannayh:
Now sorting packages alphabetically when using redis cache (it works this way on all other caches).


stevearc:
LGTM, thanks!

PR #321, opened 2022-10-01
Fix pypicloud breaking when storage.region_name = "us-east-1"

nivintw:
us-east-1 is an invalid value for location constraint. [ref](https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetBucketLocation.html) and [ref](https://docs.aws.amazon.com/AmazonS3/latest/API/API_CreateBucket.html)
The current logic means that if storage.region_name is set to "us-east-1" then the deployment will crash.

It could be argued that someone could not set this value in the config if their region were us-east-1 and it wouldn't be unfair.
However, this very minor update means that things "just work" if you always set the region_name, which IMHO is a worthwhile QOL update.


stevearc:
Yeah this seems like a clear win. Thanks for the PR!

PR #326, opened 2022-10-14
fix: add slash to stop pointless redirect

brycedrennan:
because you have `addslash` [here](https://github.com/stevearc/pypicloud/blob/master/pypicloud/views/simple.py#L109) all these urls are redirects


stevearc:
Thanks for the PR!

Issue #327, opened 2022-11-04
pypi.disallow_fallback doesn't work


pulecp:
Hello,

I would need to help with configuring `pypi.disallow_fallback`. It seems like any format doesn't work for me:

```
pypi.disallow_fallback = package1 package2
```
```
pypi.disallow_fallback =
    package1
    package2
```
```
pypi.disallow_fallback = (package1, package2)
```
```
pypi.disallow_fallback = [package1, package2]
```
```
pypi.disallow_fallback = package1, package2
```

I still see in logs that Pypicloud tried to search a package in an upstream even when it shouldn't look for it because it's private package:

```
2022-11-04 14:17:22	WARNING 2022-11-04 14:17:22,547 [pypicloud.locator] Error fetching 'package1' from upstream: 404 Client Error: Not Found for url: https://pypi.org/pypi/package1/json
```

Here is a part of configuration I'm using. If you need more or whole config, let me know.

```
pypi.always_show_upstream = true
pypi.cache_update = everyone
pypi.fallback = cache
pypi.disallow_fallback =
```

I'm running the newest version of Pypicloud from docker hub [stevearc/pypicloud:1.3.11-alpine](https://hub.docker.com/layers/stevearc/pypicloud/1.3.11-alpine/images/sha256-bbed170aeeb6fab4e074bee60129df2ee07286a046eb78aaf82e2c7c87aa6974).

I have feeling that this option doesn't work at all.

Note: I already asked the same question in an old issue where someone had similar issue with formatting - [link](https://github.com/stevearc/pypicloud/issues/85#issuecomment-1302379134).

Thanks you very much for help! If you need more info, testing, just let me know.

stevearc:
The first two formats should both work. If not, it's possible there's a bug in the logic somewhere. As per #325, I won't be working on this but since it would fall under the category of bugfix I would happily review a PR!

pulecp:
> The first two formats should both work. If not, it's possible there's a bug in the logic somewhere. As per #325, I won't be working on this but since it would fall under the category of bugfix I would happily review a PR!

You are right. Both formats are working.

There is just a wrong ordering when Pypicloud search for releases in an upstream even when the package is listed in the `pypi.disallow_fallback` option. See my PR #328 with a fix.

PR #328, opened 2022-11-09
Do not search for releases when fallback disallowed

pulecp:
The fuction "request.locator.get_releases" looking for releases in an upstream was called regardless of permissions that were set on a given package. Looking for releases of packages in the upstream that were disallowed by the "pypi.disallow_fallback" option exposed the names of disallowed packages to the upstream. This is unsolicited behavior.

This change also fixes these warnings that were logged for all disallowed packages:

WARNING [pypicloud.locator] Error fetching 'package1' from upstream: 404 Client Error: Not Found for url: https://pypi.org/pypi/package1/json

Fixed issue #327

### DIFF
--- a/pypicloud/views/simple.py
+++ b/pypicloud/views/simple.py
@@ -161,10 +161,10 @@ def package_versions_json(context, request):
 
 def get_fallback_packages(request, package_name, redirect=True):
     """Get all package versions for a package from the fallback_base_url"""
-    releases = request.locator.get_releases(package_name)
     pkgs = {}
     if not request.access.has_permission(package_name, "fallback"):
         return pkgs
+    releases = request.locator.get_releases(package_name)
     for release in releases:
         url = release["url"]
         filename = posixpath.basename(url)


stevearc:
Change looks good, but there appear to be some issues with the CI tests. Could you look into it?

pulecp:
I would like to help but it seems like the CI is broken. See the error. I can't be caused by my change:

```
py39 run-test: commands[0] | coverage run --source=pypicloud --branch setup.py nosetests --no-skip
Traceback (most recent call last):
  File "/usr/lib/python3.9/pathlib.py", line 387, in gethomedir
    return os.environ['HOME']
  File "/usr/lib/python3.9/os.py", line 679, in __getitem__
    raise KeyError(key) from None
KeyError: 'HOME'
```

stevearc:
While I understand that your change didn't cause the CI error, I do not have the time to look into it myself. I put pypicloud in maintenance mode because I have too many other demands on my time, and I can't spend a random hour digging into CI failures on a project I don't use anymore.

pulecp:
> While I understand that your change didn't cause the CI error, I do not have the time to look into it myself. I put pypicloud in maintenance mode because I have too many other demands on my time, and I can't spend a random hour digging into CI failures on a project I don't use anymore.

I see and understand that you have different priorities now. Let's keep this PR open so other can see it. It might be accepted one day :).
Thank you for your time you invested into this project. I still hope that someone will over this project and will continue in maintenance and further development. Let's see!

PR #330, opened 2023-07-18
Set pypi.default_read = authenticated in ppc-make-config

ddelange:
Hi 👋 

We found out that our kubernetes cluster was exposing packages due to `default_read=everyone`. We created the config for our cluster using the 'docker mode' (`ppc-make-config -r`). This default was unexpected to us, also because of the [docs](https://pypicloud.readthedocs.io/en/latest/topics/configuration.html#pypi-default-read) stating `authendicated` as default value.

### DIFF
--- a/pypicloud/templates/config.ini.jinja2
+++ b/pypicloud/templates/config.ini.jinja2
@@ -8,7 +8,7 @@ pyramid.debug_routematch = false
 pyramid.default_locale_name = en
 
 pypi.default_read =
-    {% if env == 'prod' -%}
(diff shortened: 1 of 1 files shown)


stevearc:
The docs are talking about what the default value is when _not_ specified by the configuration file. I can understand missing this if you just generate the config file and deploy it, but I think anyone reading over the config, especially if looking specifically at the ACL, would be able to spot it.

ddelange:
alright, fair enough:) thanks for the quick reply

